PR #1, opened 2025-10-08
Fix streaming completion race and add reproduction script

vikhyat:
## Summary
- fix the EngineStream completion signalling race by scheduling the sentinel on the event loop
- add a small async script that reproduces the original ordering bug and demonstrates the fix

## Testing
- python examples/repro_stream_race.py

------
https://chatgpt.com/codex/tasks/task_e_68e5ab465320832b9659be0ea4aa9bc7

### DIFF
--- a/examples/repro_stream_race.py
+++ b/examples/repro_stream_race.py
@@ -1,0 +1,61 @@
+"""Demonstrate the streaming completion race in the inference engine."""
+
+from __future__ import annotations
+
+import asyncio
+
+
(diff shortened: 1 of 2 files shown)


vikhyat:
we don't want to merge the test script, that's just for verifying the issue is real and local testing. need a new PR with just the fix, once we've confirmed the issue is real.

vikhyat:
@codex can self._loop actually be None where we added that guard? or were we being overly cautious?

vikhyat:
Fixed by #2

PR #4, opened 2025-10-09
Fix KV cache writes and stabilise MoE CUDA graph replay

vikhyat:
## Summary
- fix element-wise KV cache writes so concurrent requests no longer trample each other's KV pages
- keep ScatterMoE CUDA graph captures in sync by recomputing padded block indices each step and removing torch.compile/custom-op paths that broke capture
- restore HTTP ramp token throughput reporting by reading the newer input/output token metrics

## Testing
- ssh p2 'cd ~/code/kestrel && PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True ~/.local/bin/uv run python examples/consistency_probe.py --weights ~/code/moondream/model.pt --question "Describe the scene in the image." --image tmp/sample.jpg --concurrency 8 --rounds 40 --max-new-tokens 256 --max-batch-size 8 --max-seq-length 8192'
- ssh p2 'cd ~/code/kestrel && PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True ~/.local/bin/uv run python examples/benchmark_http_server.py --url http://127.0.0.1:8080/v1/query --image external/moondream/assets/demo-1.jpg --stage-duration 30 --start-concurrency 1 --concurrency-step 4 --max-concurrency 64 --output tmp/http_ramp.json'
- ssh p2 'cd ~/code/kestrel && python3 - <<"PY"
import base64, json, urllib.request
from pathlib import Path
image_path = Path("external/moondream/assets/demo-1.jpg")
data = base64.b64encode(image_path.read_bytes()).decode()
payload = {
    "question": "Describe the scene in the image.",
    "image_url": "data:image/jpeg;base64," + data,
    "settings": {
        "temperature": 0.0,
        "top_p": 1.0,
        "max_tokens": 128
    }
}
req = urllib.request.Request(
    "http://127.0.0.1:8080/v1/query",
    data=json.dumps(payload).encode("utf-8"),
    headers={"Content-Type": "application/json"},
)
with urllib.request.urlopen(req, timeout=60) as resp:
    print(resp.status)
    body = resp.read().decode("utf-8")
    print(body)
PY'

### DIFF
--- a/examples/benchmark_http_server.py
+++ b/examples/benchmark_http_server.py
@@ -318,7 +318,11 @@ def aggregate_token_stats(
     success = 0
     for rec in records:
         prompt_val = rec.server_metrics.get("prompt_tokens")
+        if not isinstance(prompt_val, (int, float)):
+            prompt_val = rec.server_metrics.get("input_tokens")
         decode_val = rec.server_metrics.get("decode_tokens")
+        if not isinstance(decode_val, (int, float)):
+            decode_val = rec.server_metrics.get("output_tokens")
         if isinstance(prompt_val, (int, float)):
             prompt_total += float(prompt_val)
         if isinstance(decode_val, (int, float)):
--- a/examples/benchmark_http_server.py
+++ b/examples/benchmark_http_server.py
@@ -317,12 +317,8 @@ def aggregate_token_stats(
     decode_total = 0.0
     success = 0
     for rec in records:
-        prompt_val = rec.server_metrics.get("prompt_tokens")
-        if not isinstance(prompt_val, (int, float)):
-            prompt_val = rec.server_metrics.get("input_tokens")
-        decode_val = rec.server_metrics.get("decode_tokens")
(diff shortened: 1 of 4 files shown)


vikhyat:
@codex review, and also i don't like the capturing checks just for running assertions, let's get rid of those

examples/benchmark_http_server.py
```diff
@@ -318,7 +318,11 @@ def aggregate_token_stats(
     success = 0
     for rec in records:
         prompt_val = rec.server_metrics.get("prompt_tokens")
+        if not isinstance(prompt_val, (int, float)):
+            prompt_val = rec.server_metrics.get("input_tokens")
         decode_val = rec.server_metrics.get("decode_tokens")
+        if not isinstance(decode_val, (int, float)):
+            decode_val = rec.server_metrics.get("output_tokens")
```
vikhyat: @codex we don't need to support the old token count metrics. Just switch to the new one.

PR #5, opened 2025-10-24
Add segmentation support to inference

vikhyat:
Not ready for merge, needs to be reviewed.


vikhyat:
@codex review

EthanReid:
This works but it has some oddities.

1: Why do we return text, svg_path, path_tokens, and token_ids, that is a lot of duplicate data.

2: We are not returning an SVG, but an SVG path, is it the user/Stations responsibility to convert to image space and actually build the SVG?

3: Should we have the prompt be called "object" instead of "label", that way it is consistent with point and detect?

EthanReid:
Should we stop processing if it generates an invalid SVG path, currently it does not check for this?

EthanReid:
TODO: come back and clean this up. Will work for RL for now. 

Note: currently only works if you add the text.

PR #11, opened 2025-12-05
Update segmentation refiner model

EthanReid:
Crude Claude-ish code to add the HQSam style head to Kestrel. Will follow up with a cleaner less verbose/more efficient implementation.

Note, the new refiner brings us from 86.9% on RefCOCO-M to 88.2%.
Note: A lot of the changes are from black formatting.
Note: **Do not delete branch, I want to look at the commits for a bit.**


kestrel/moondream/vision.py
```diff
@@ -64,6 +64,27 @@ def vision_encoder(crops: torch.Tensor, module: nn.Module, config: VisionConfig)
     return x
 
 
+def vision_encoder_multiscale(
```
EthanReid: Note: VE now can return an early feature and the actual vision encoding. We use the early feature in the mask refinement. This idea was lifted directly from HQSam.

vikhyat:
@codex review

PR #12, opened 2026-01-18
spatial query

EthanReid:
Spatial Query was missing from Kestrel.

Added spatial query support by accepting `spatial_refs` on the query API/engine path, validating them alongside the image requirement, and threading them into the QuerySkill request. Query prompt construction now inserts Coord/Size tokens for point and bbox refs in front of the question text using the same format as segment.

This is a cleaned up version of what I have been using for my spatial fine tunes. I tested this PR by running some bbox and point queries.

### DIFF
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -29,7 +29,6 @@ import asyncio
 import hashlib
 import itertools
 import logging
-import math
 import queue
 import threading
 import time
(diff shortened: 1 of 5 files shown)


EthanReid:
@codex review

EthanReid:
@codex review

kestrel/engine.py
```diff
@@ -880,6 +859,11 @@ def _normalize_adapter_id(self, value: Optional[str]) -> Optional[str]:
             raise ValueError("adapter must be a non-empty string")
         return normalized
 
+    def _normalize_spatial_refs(
+        self, spatial_refs: Optional[Sequence[Sequence[float]]]
+    ) -> Optional[Tuple[Tuple[float, ...], ...]]:
+        return normalize_spatial_refs(spatial_refs)
+
```
vikhyat: wrapper seems unnecessary, why not just call directly?

EthanReid: Was a wrapper to be consistent with _normalize_adapter_id, _normalize_top_p, ect... 

Directly calling it is a functionally the same, updated the PR.

PR #15, opened 2026-02-12
Ethan/md3 prompt fix

EthanReid:
When MD2 was added, a special gate was added to add the BOS token to the start of the prompt, citing MD2 hugging face behavior. The issue is that both MD2 and MD3 have the same behavior here in HF. This shows a bug in the MD3 kestrel logic, mode_id is being placed before the image, not BOS. This is as kestrel takes the first token in the prompt and places it behind the image, but in the MD3 path, there is no BOS being added, so that token is instead mode_id.

[MD2 Code](https://huggingface.co/vikhyatk/moondream2/blob/main/moondream.py#L569)
[MD3 Code](https://huggingface.co/moondream/moondream3-preview/blob/main/moondream.py#L656)

Note: this is deployed in staging, running the latest kestrel and kestrel-kernel.

### DIFF
--- a/kestrel/skills/caption.py
+++ b/kestrel/skills/caption.py
@@ -58,9 +58,7 @@ class CaptionSkill(SkillSpec):
                 f"Unsupported caption length '{length_key}'. Expected one of: {valid}"
             )
         token_ids = templates[length_key]
-        tokens = []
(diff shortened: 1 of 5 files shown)


EthanReid:
@codex review

PR #20, opened 2026-04-23
Use selected FP8 Triton MoE config

vikhyat:
Summary:
- Route the FP8 Triton MoE launch through the prepared triton_config instead of the global default.
- Keeps routing block size and Triton launch config consistent for tuned configs.

Validation:
- Modal RTX PRO 6000 smoke with latest kestrel-kernels nightly wheel plus m87-labs/kestrel-proprietary#30 overlay: passed, ChartQA limit 8, batch size 64, direct mode.
- CUDA_LAUNCH_BLOCKING=1 debug run moved the previous illegal access to a config OOR, and this change plus the kernels-side fallback/config overlay cleared it.

Related:
- Requires m87-labs/kestrel-proprietary#30 for RTX 6000 kernel config packaging.

### DIFF
--- a/kestrel/fused_moe/module.py
+++ b/kestrel/fused_moe/module.py
@@ -827,7 +827,7 @@ def _invoke_fused_moe_kernel(
             num_tokens_post_padded=num_tokens_post_padded,
             mul_routed_weight=mul_routed_weight,
             top_k=top_k,
-            config=_TRITON_FP8_CONFIG,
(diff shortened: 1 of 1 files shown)


vikhyat:
Closing this as obsolete. Current main has refactored the MoE launch path through kestrel-kernels runtime dispatch, so this stale branch conflicts and is no longer the right fix.

PR #23, opened 2026-04-23
Gate MoE-LoRA kernels on triton availability

vikhyat:
## Summary
Make ``kestrel.fused_moe`` importable on platforms without NVIDIA's ``triton`` package (Windows, some Jetson images). Moondream3 base inference is blocked today because ``kestrel/fused_moe/lora_kernels.py`` does a hard ``import triton`` at module load — triggered via ``build_moe_mlp`` → ``FusedMoEModule`` on every engine init, even with no LoRA adapters attached.

## What changes
- Move the two ``@triton.jit`` kernels into a new ``_lora_triton_kernels`` submodule. Only the LoRA code path pulls it in.
- In ``lora_kernels.py``, wrap ``import triton`` in try/except and set ``_TRITON_AVAILABLE``. Module now imports on any platform.
- ``apply_moe_lora_batched`` / ``apply_moe_lora_single`` raise ``RuntimeError`` at call time when ``_TRITON_AVAILABLE`` is False, with a message pointing at the platform constraint.

## Why not a plain-torch fallback
I drafted one (batched gather + ``bmm``), but the kernels run inside CUDA graph capture. A torch equivalent needs fixed-shape operations with masked scatters and no host syncs — about 150 LoC of careful mask/gather/scatter plumbing that obscures the semantics. Gating is cleaner as a first step; LoRA adapters are unavailable on Windows/Jetson-lite today but base moondream2 / moondream3 inference works. If we need plain-torch LoRA later, it drops in behind the same ``_TRITON_AVAILABLE`` switch.

## Test plan
- [x] On Windows L4 (no triton): ``from kestrel.fused_moe import FusedMoEModule`` succeeds, ``apply_moe_lora_batched`` raises a clean RuntimeError on call.
- [x] Moondream2 ChartQA smoke on Windows L4 (10 samples, batch=4): 10.54 req/s, 357 ms P50 — matches pre-change numbers.
- [ ] On Linux (triton present): full test suite; ChartQA matches baseline. Haven't re-run yet — happy to if the reviewer wants, but the Triton code path is byte-for-byte unchanged (just moved to a submodule).

### DIFF
--- a/kestrel/fused_moe/_lora_triton_kernels.py
+++ b/kestrel/fused_moe/_lora_triton_kernels.py
@@ -1,0 +1,278 @@
+"""Triton MoE-LoRA kernels — imported only when the `triton` package is
+available. Kept out of ``lora_kernels.py`` so the top-level module can
+be imported in Triton-less environments (Windows, some Jetson images)
+and transparently delegate to a plain-torch fallback."""
+
+
+import triton
+import triton.language as tl
+
+
+@triton.jit(
+    do_not_specialize=[
+        "EM",
+        "num_valid_tokens",
+        "max_tokens_padded",
+        "stride_tl",
+        "stride_el",
+    ]
+)
+def _batched_moe_lora_kernel(
+    # Input/output pointers
+    a_ptr,  # Input: [num_valid_tokens, K] - x for shrink, intermediate for expand
+    b_ptr,  # Weights: [max_loras * num_experts, N, K] - lora_a or lora_b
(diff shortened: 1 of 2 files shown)


vikhyat:
Superseded: moving the Triton kernels into kestrel-kernels where other kernels live — much smaller diff on the kestrel side, platform gating done at the kernel boundary. Will re-open once the kestrel-kernels side is staged.

PR #29, opened 2026-04-24
Avoid redundant prefill seqused_q

vikhyat:
## Summary

Only pass `fa3_seqused_q` from runtime prefill when the Q batch is actually padded. Uniform-Q prefill, including the single-token prefix-cache append case, now passes `None` so the existing fixed-Q/paged fast paths can dispatch normally.

This keeps `seqused_q` reserved for its intended meaning: per-row valid Q lengths for a padded varlen-Q batch.

## Why

The full ChartQA dispatch probe showed one warmup/cache-reuse text forward falling through to torch SDPA on L40S/SM89. Root cause: runtime passed `seqused_q=[1]` for a single-token append prefill. That redundant varlen marker blocked the existing fast path even though every row had the full Q length.

## Validation

- `python -m py_compile kestrel/moondream/runtime.py tests/moondream/test_runtime_prefill.py`
- `.venv/bin/python -m pytest tests/moondream/test_runtime_prefill.py` -> `3 passed`
- `git diff --check`
- Modal L40S / SM89 `ap-3fEIT8u3JPoYfzJsJTjW3I`, Moondream 2 ChartQA `--limit 1`: completed successfully with dispatch counts `{"decode_persistent_split_fused": 48, "sm89_decode": 744, "sm89_paged_prefill": 24, "sm89_prefill": 48}` and no `torch_sdpa_fallback`.

### DIFF
--- a/kestrel/moondream/runtime.py
+++ b/kestrel/moondream/runtime.py
@@ -1426,7 +1426,8 @@ class MoondreamRuntime:
 
             assert use_prefix_attn is not None
             hidden_dim = self.bos_embed.shape[-1]
-            max_seq_len = max(t.shape[1] for t in per_inputs)
+            seq_lens = [int(t.shape[1]) for t in per_inputs]
+            max_seq_len = max(seq_lens)
             inputs_embeds = torch.zeros(
                 (batch_size, max_seq_len, hidden_dim),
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

PR #31, opened 2026-04-26
Drop kestrel/ops/layernorm_cuda; route through DenseRuntime directly

vikhyat:
## Summary
Routing debt cleanup paired with **kestrel-proprietary#79**. The high-level layernorm API (shape validation, variant selection, fallback-to-torch) now lives in kestrel-kernels' ``DenseRuntime`` — kestrel just calls ``_KERNELS.dense.layernorm_bias_into(...)`` directly, like every other op (rotary, tau, flash_attn).

The MPS Metal LayerNorm kernel from kestrel-proprietary#71 was sitting completely unused on MPS today: ``layers.py:40`` and ``vision.py:79`` both gated the fast path on ``x.is_cuda and x.dtype == torch.bfloat16``, so fp16-on-MPS fell through to ``F.layer_norm``. After this pair lands, the Metal kernel actually runs.

## Changes
- **Delete** ``kestrel/ops/layernorm_cuda/`` entirely. The wrapper duplicated kernel-API responsibilities that belong in kestrel-kernels.
- **``moondream/layers.py``**: drop the ``is_cuda + bf16`` gate; always call ``_KERNELS.dense.layernorm_bias`` with ``fallback_to_torch=True``.
- **``moondream/vision.py``**: same fix in the vision-encoder ``_layer_norm``. Drop ``use_fast_ln`` and its ``is_cuda + bf16 + not autograd`` conditions — kestrel is inference-only, autograd is never enabled.
- **``ops/fused_mlp/__init__.py``**: internal alias renamed (``..._cuda`` → ``..._kernel``) to match the new ``DenseRuntime`` field name. No behavior change.
- **``scripts/chartqa_eval.py``**: add ``--device`` flag with auto-detect (defaults to MPS on a torch build without CUDA, with ``enable_cuda_graphs=False``). The auto-detect existed in a prior Mac install but wasn't in main; this restores it so chartqa_eval runs out of the box on Mac.

## ChartQA on M2 (--limit 30, --max-batch-size 1, --kv-cache-pages 1024, default prefix cache)

vs prior all-fp16 baseline (kestrel-kernels with #71/#72/#73/#74/#76/#77 + kestrel#30):

| metric | prior | **this PR** | Δ |
|---|---|---|---|
| Accuracy | 31/60 | 30/60 | within sampling noise |
| Prefill | 397.02 tok/s | **442.34** | **+11%** |
| Decode | 1.39 tok/s | **1.57** | **+13%** |
| Requests/s | 0.52 | **0.58** | +12% |
| P50 latency | 2037 ms | **1932** | −5% |
| P90 latency | 3844 ms | **3616** | −6% |
| P99 latency | 4311 ms | **3949** | −8% |

Cumulative MPS speedup since the original baseline (across LayerNorm + RoPE + tau + flash_attn_dense + paged decode + fp16 + this routing fix): **prefill +214%, decode +220%, P99 −76%**.

## Pairing
Requires **kestrel-proprietary#79** (renames the ``DenseRuntime`` field to ``layernorm_bias_into`` / ``layernorm_bias`` and adds the high-level API to each backend). These two PRs need to merge together — kestrel will fail at import if it goes first, and kestrel-kernels has nothing calling the new fields without this PR.

## Test plan
- [x] Smoke: ``_KERNELS.dense.layernorm_bias`` resolves to the MPS impl on Mac, output matches ``F.layer_norm`` at ~1 ULP for fp16
- [x] ChartQA --limit 30 end-to-end on M2 (table above)
- [x] No fallback warnings beyond the expected paged/prefix_lm_730 (still on torch SDPA pending the prefix LM prefill kernel — separate follow-up)
- [ ] CUDA path: smoke test on an A100 once the paired PR lands


vikhyat:
@codex review

vikhyat:
@codex review

PR #33, opened 2026-04-27
engine: warn once when sampling settings exit the MPS kernel envelope

vikhyat:
## Summary

The MPS fused ``sample_step`` kernel (kestrel-proprietary #84) only
inspects the top-K=64 candidates and is calibrated for ``top_p ≤
0.95`` / ``temperature ≤ 1.0``. Settings outside that envelope can
produce biased samples:

- **High top_p (≥ 0.95)**: the kernel truncates at K=64 even when
  the true cumulative-mass crossing is later, biasing the nucleus
  toward the head.
- **High temperature (> 1.0)**: the post-temp distribution flattens
  enough that K=64 may not span the support.

We accept the trade-off (full kernel speed — ~5× the torch chain
on V=51200) but surface a **one-time per-process** warning the
first time a request with out-of-envelope settings is admitted on
an MPS runtime. Users running with non-default sampling settings
know their output distribution will diverge from the CUDA / torch
reference.

CUDA / CPU runtimes are unaffected — the warning gate keys on
``runtime.device.type == "mps"``.

## Implementation

- ``InferenceEngine._warn_if_outside_mps_sampler_envelope`` — gates
  on ``runtime.device.type``, dedupes via a class-level
  ``set[(temp, top_p)]`` so a long chat session doesn't spam.
- Called from ``_submit_request`` right after the existing
  ``_normalize_temperature`` / ``_normalize_top_p`` validation, so
  every entry point (query, caption, detect, ask, raw submit) is
  covered.

## Test plan

- [x] Syntax check
- [ ] Manual: on Mac, run with ``settings={'top_p': 0.99}`` and
      verify the warning fires once
- [ ] Manual: same call again with same settings → no second
      warning

### DIFF
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -88,6 +88,11 @@ from kestrel.utils.spatial_refs import normalize_spatial_refs
 
 
 _LOGGER = logging.getLogger(__name__)
+
+# PR landing this warning + the kernel envelope it documents. Update if
+# the canonical docs move.
+_MPS_SAMPLER_DOCS_URL = "https://github.com/m87-labs/kestrel/pull/33"
+
 _DEFAULT_API_BASE_URL = "https://api.moondream.ai"
 _ALLOWED_API_BASE_URLS = (
     "https://api-staging.moondream.ai",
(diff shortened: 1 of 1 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

PR #41, opened 2026-04-30
release: kestrel 0.3.0 (Apple Silicon + Blackwell + Windows + Jetson Thor)

vikhyat:
## Summary
- Switch the global-crop resize in \`image_crops.py\` from out-of-place \`kestrel_native.resize_lanczos3\` to the new in-place \`kestrel_native.resize_lanczos3_into\` (added in kestrel-native 0.1.4).
- Bump pin: \`kestrel-native==0.1.3\` → \`kestrel-native==0.1.4\`.

## Why
\`image_crops\` already builds a single contiguous \`np.empty((num_crops, H, W, 3))\` buffer for all crops. The global crop was the one slot that still went through the out-of-place API, which allocates a fresh ~440 KB tensor inside the native extension and then copies it into \`crops[0]\` via numpy assignment. The in-place variant validates contiguity / writability / src-dst aliasing, and \`crops[0]\` (a contiguous view into the pre-allocated buffer) satisfies all three. This eliminates one allocation + memcpy per request.

The tiled-resize path is unchanged — its output is read repeatedly to slice tile crops, with no pre-allocated destination in scope.

## Test plan
- [ ] \`pytest\` (or a targeted run on \`tests/moondream/\` if there's a fast subset) passes.
- [ ] Manual smoke: ChartQA eval at \`limit=10\` returns the same answers as a pre-PR build (no numerical regression in the resize result).


vikhyat:
@codex review

vikhyat:
Addressed codex P1: bumped `kestrel-kernels` 0.2.1 → 0.3.0 (which now ships Windows x86_64 + macOS arm64 wheels alongside Linux). Lock no longer needs the `sys_platform == 'linux'` marker — verified with `uv sync --frozen --dry-run --python-platform {linux,windows}` (both pass). macOS dry-run errors against `macosx_13_0` because the kestrel-kernels CI uses the `macos-14` runner so the wheel is tagged `macosx_14_0_arm64`; real installs on macOS 14+ Python 3.12 resolve fine.

@codex review

vikhyat:
Pushed `bc8c3b8` to address the macOS angle. Added `tool.uv.environments` scoping the supported set to:

```
sys_platform == 'linux'
sys_platform == 'win32'
sys_platform == 'darwin' and platform_machine == 'arm64' and python_version == '3.12'
```

This matches what kestrel-kernels actually ships: Linux x86_64/aarch64 and Windows x86_64 cover Python 3.10–3.13; macOS arm64 ships cp312 only (the kestrel-kernels CI uses the `macos-14` GitHub runner so the wheel tag is `macosx_14_0_arm64`).

Pre-PR, `uv sync --python-platform aarch64-apple-darwin` 'succeeded' purely because the stale `sys_platform == 'linux'` marker silently skipped kestrel-kernels on macOS — but kestrel needs kestrel-kernels at runtime on macOS for the MPS Metal kernels (rotary, paged decode/prefill, MoE, …), so install would have crashed at first use. With `tool.uv.environments` declared, uv now fails at sync time with a clear 'platform not compatible' error on macOS-arm64 + Python 3.10/3.11/3.13 — which is the honest answer.

The remaining edge: `uv sync --frozen --dry-run --python-platform aarch64-apple-darwin --python 3.12` still errors against `macosx_13_0` (uv's hardcoded default) even though the wheel is `macosx_14_0`. That's a real-world non-issue — actual macOS 14+ users have a higher target tag and resolve cleanly — but a kestrel-kernels CI follow-up to also publish a `macosx_13_0` (or `macosx_11_0`) tagged wheel would close it. Not in scope for this PR.

@codex review

vikhyat:
Pushed `5c7d9e4` reverting the platform_machine constraints from `bc8c3b8`.

I tried codex's P2 suggestion (tighten `platform_machine` so unsupported arches like Windows ARM64 / Linux i686 fail fast). It collides with two uv-default-target quirks:

- **Linux aarch64**: our wheels are tagged `manylinux_2_34/_2_35`, but uv's `--python-platform aarch64-unknown-linux-gnu` targets `manylinux_2_28` by default. With `platform_machine == 'aarch64'` the dry-run fails ("no compatible distribution") even though real Ubuntu 22.04 / RHEL 9 (glibc ≥ 2.34) installs resolve cleanly.
- **Windows x86_64**: `platform.machine()` returns `AMD64` at runtime but uv normalizes `x86_64-pc-windows-msvc` to `x86_64` at lock time, so neither value matches both contexts cleanly.

OS-level scope is the practical floor here — real users on supported arches resolve cleanly, and codex's P2 nit (late vs early failure on truly unsupported arches like Windows ARM64) is one level later, not a correctness regression. Documented in the PR commit + a comment block in pyproject.toml.

For codex's P1 (still flagged on `bc8c3b8` and now `5c7d9e4` — `uv sync --python-platform aarch64-apple-darwin` baseline-macos-13 dry-run fails because our macOS wheel is `macosx_14_0_arm64`): this is a uv tooling artifact rather than a real install regression. The proper fix is on the kestrel-kernels CI side — switching the macos-14 runner to macos-13, or also publishing a `macosx_11_0` / `macosx_13_0` tagged wheel. Tracking that as a follow-up; not in scope for this PR.

@codex review

vikhyat:
Closing the codex macOS-13 P1: kestrel-kernels 0.3.1 lowered the macOS arm64 wheel's deployment target to 13.0 (Ventura) by dropping bf16 from the MPS surface. Pushed `7735670` bumping the pin 0.3.0 → 0.3.1.

Verified with `uv sync --frozen --dry-run --python-platform`:
- `x86_64-unknown-linux-gnu` ✓
- `x86_64-pc-windows-msvc` ✓
- `aarch64-apple-darwin --python 3.12` (uv's default macos-13 target) ✓ — **this is the case 0.3.1 unblocks**.
- `aarch64-apple-darwin` (default Python 3.13, unsupported by our cp312-only macOS wheel) → fails fast with "platform not compatible with the lockfile's supported environments" via `tool.uv.environments`.

Production kestrel was already running fp16 on MPS via `kestrel/config.py: resolved_dtype`, so the 0.3.1 bf16 drop is a no-op at runtime. CUDA paths unaffected.

@codex review

vikhyat:
Pushed `dd937b5` with the 0.3.0 release: version bump 0.2.1 → 0.3.0, regenerated `uv.lock`, and a customer-facing CHANGELOG entry covering the four new platforms (Apple Silicon, Windows, Blackwell, Jetson Thor) plus the cross-arch perf wins on existing NVIDIA hardware.

Renamed the PR title to reflect that this is now the 0.3.0 release commit, not just the in-place-resize adoption.

@codex review

vikhyat:
Three doc/dep updates pushed:

- `7b93954` — refresh README.md (NVIDIA SM list extended through Blackwell + Apple Silicon path called out, Metal mentioned alongside CUDA in the lead, Jetson pointer covers both Orin / JP6 and Thor / JP7) and rewrite `docs/jetson.md` to be parallel for Orin (custom NVIDIA torch wheel + JetPack /usr/local/cuda paths) and Thor (standard PyPI torch + venv-resident nvidia-cu13 + nvpl libs). Verified the Thor install + LD_LIBRARY_PATH instructions on the Thor box: `torch.cuda.is_available()` True, kestrel imports clean.
- `76b859c` — drop `torch<2.11` cap. Upstream pytorch issue #178255 (the reason for the cap) is closed; the bug was specific to Python 3.13.8 + torch 2.11.0 and doesn't reproduce on Python 3.10/3.11/3.12. Lock regenerated.

@codex review

vikhyat:
Pushed `5b634f1` — drop `apache-tvm-ffi` from kestrel runtime deps. kestrel itself never imports it; kestrel-kernels 0.3.1 dropped tvm-ffi from its runtime metadata (it's now build-system.requires + a PEP-735 dependency-group only, neither of which lands in the wheel). Net effect: smaller install, 31 wheel entries removed from uv.lock.

@codex review

PR #47, opened 2026-05-05
Centralize per-layer KV allocation behind a KVMemoryPool

vikhyat:
## Summary

Funnel every \`PagedKVCache\` K/V buffer allocation through a new \`KVMemoryPool\` so storage isn't bound to \`PagedKVCache.__init__\`. The pool is currently a thin wrapper around \`torch.zeros\` that also reports \`allocated_bytes\` for diagnostics; the abstraction exists so:

- future revisions can share an underlying buffer or implement compaction without touching call sites
- the engine can eventually host multiple runtimes that share the GPU's KV budget
- KV memory accounting has a single chokepoint today

## Changes

- New \`KVMemoryPool\` in \`kestrel/kv_cache.py\`. Constructed with a \`device\`; exposes \`allocate_layer_kv(n_pages, n_heads, page_size, head_dim, dtype) -> (k, v)\` and tracks total bytes in \`allocated_bytes\`.
- \`PagedKVCache\` now requires a \`pool\` parameter; it asks the pool for its K and V buffers instead of calling \`torch.zeros\` directly.
- \`MoondreamRuntime\` constructs one \`KVMemoryPool\` per runtime instance and passes it to every \`_LayerPagedCache\`. \`_LayerPagedCache\` no longer takes \`device\` (the pool already knows it), and the redundant \`.to(device)\` on the inner \`PagedKVCache\` is gone.

## Test plan
- [x] \`pytest --ignore=tests/integration\` — 242 passed, 46 skipped, 0 changed
- [x] Smoke-tested pool wiring: \`PagedKVCache\` allocations land on the pool's device; \`allocated_bytes\` matches \`2 × n_pages × n_heads × page_size × head_dim × element_size\` exactly.
- [x] No new GPU allocations vs. baseline — same per-layer \`torch.zeros\` calls, just routed through \`KVMemoryPool\`.

### DIFF
--- a/kestrel/kv_cache.py
+++ b/kestrel/kv_cache.py
@@ -129,10 +129,12 @@ class PagedKVCache(torch.nn.Module):
             self.k_scale = 1.0
             self.v_scale = 1.0
         self.register_buffer(
-            "k_scale_tensor", torch.tensor(self.k_scale, dtype=torch.float32)
+            "k_scale_tensor",
+            torch.tensor(self.k_scale, dtype=torch.float32, device=pool.device),
         )
         self.register_buffer(
-            "v_scale_tensor", torch.tensor(self.v_scale, dtype=torch.float32)
+            "v_scale_tensor",
+            torch.tensor(self.v_scale, dtype=torch.float32, device=pool.device),
         )
 
     def update(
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

PR #48, opened 2026-05-05
Type the scheduler against a Runtime protocol

vikhyat:
## Summary

Introduce \`kestrel.runtime.protocol.Runtime\` — a Protocol declaring the methods and properties the \`GenerationScheduler\` calls on a runtime. Annotate the scheduler against it so the scheduler/runtime contract is explicit at the type level and so the scheduler isn't bound to \`MoondreamRuntime\` as the only valid runtime type.

This is groundwork for hosting multiple model runtimes under one engine; the scheduler is the cleanest call site to start with (it only invokes runtime methods, never reaches into internal state).

## Approach

Model-specific parameter types (image crops, adapters, prefill/decode slots) are typed as \`Any\` in the protocol; concrete runtimes narrow them. This avoids embedding Moondream-specific types (\`OverlapCropOutput\`, \`LoRA\`, \`PrefillSlot\`, \`DecodeSlot\`) into the protocol surface.

## Changes

- new \`kestrel/runtime/protocol.py\` defining the \`Runtime\` protocol
- \`kestrel.runtime\` re-exports \`Runtime\`
- \`kestrel/scheduler/scheduler.py\` and \`kestrel/scheduler/types.py\` replace their \`MoondreamRuntime\` imports + parameter annotations with \`Runtime\`

## Out of scope

The engine still holds a concrete \`MoondreamRuntime\` — it reaches into runtime internals like \`.config.vision\` and \`.active_sequences\`. Narrowing the engine's surface so it can be typed against \`Runtime\` is a follow-up.

## Test plan
- [x] \`pytest --ignore=tests/integration\` — 242 passed, 46 skipped, 0 changed
- [x] \`pyright kestrel/scheduler kestrel/runtime kestrel/engine.py\` — 0 errors. \`MoondreamRuntime\` structurally satisfies \`Runtime\`; the engine's \`GenerationScheduler(runtime=...)\` construction site still type-checks.

### DIFF
--- a/kestrel/runtime/__init__.py
+++ b/kestrel/runtime/__init__.py
@@ -2,10 +2,12 @@
 
 This package collects the model-agnostic dataclasses, named tuples, and
 helper types that the engine and scheduler exchange with a runtime
-implementation. Concrete runtimes (e.g. ``kestrel.moondream.runtime``)
-build on top of these.
+implementation, plus the :class:`Runtime` protocol describing the
+scheduler↔runtime contract. Concrete runtimes (e.g.
+``kestrel.moondream.runtime``) build on top of these.
 """
 
+from kestrel.runtime.protocol import Runtime
 from kestrel.runtime.state import (
     PrefillClassification,
     PreparedSequence,
@@ -23,6 +25,7 @@ __all__ = [
     "CoordToken",
     "PrefillClassification",
     "PreparedSequence",
+    "Runtime",
     "RuntimeDecodeResult",
     "SequenceState",
     "SizeToken",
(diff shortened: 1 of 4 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

Updated since the previous review: `fe752d2` adds the missing attrs (`config`, `region`, `spatial_tables`, `page_table`, `device`, `max_batch_slots`, `primary_stream`, `copy_stream`, `prefill_slots`, `decode_slots`, `active_sequences`) to the Protocol. Model-agnostic ones are narrowly typed; Moondream-specific ones are `Any` with a comment that narrowing or refactoring the scheduler away from them is a follow-up.

PR #49, opened 2026-05-05
Add a FakeRuntime that satisfies the Runtime protocol for scheduler tests

vikhyat:
## Summary

The \`GenerationScheduler\` is now typed against \`kestrel.runtime.Runtime\` (introduced in #48), but the scheduler test suite only had a partial 4-method ad-hoc \`_RuntimeStub\`. This PR adds a complete \`FakeRuntime\` that satisfies the full protocol surface, plus a structural-conformance test that pins them together.

Running the fake gives the protocol a second consumer, validating that the surface in #48 is actually satisfiable by something other than \`MoondreamRuntime\`.

## What's added

**\`tests/scheduler/_fake_runtime.py\` — \`FakeRuntime\`**:
- attributes default to plausible test-shape values (\`max_batch_size=1\`, \`page_size=1\`, etc.); mutable so individual tests can override
- methods record their arguments into public lists (\`acquired_prefill_slots\`, \`released_sequences\`, \`launch_calls\`, …) so tests can assert ordering and arguments
- \`prepare_sequence\` and \`launch_prepared_batch\` accept a pre-built return value or an exception so failure-path tests can pin behaviour without constructing a real \`PreparedSequence\`
- \`_FakePageTable\` stand-in exposes \`page_size\` + \`commit_block_table\` no-op so decode-step tests can run

**\`tests/scheduler/test_runtime_conformance.py\`**:
- walks \`Runtime\`'s declared annotations + methods and asserts each one is present on \`FakeRuntime\`, so future protocol changes can't silently drift the fake out of sync

## What's migrated

\`tests/scheduler/test_launch_prefill_step.py\` previously defined a 4-method \`_RuntimeStub\` inline. Replaced with \`FakeRuntime\`; the parameterised test still passes.

## Test plan
- [x] \`pytest --ignore=tests/integration\` — 243 passed (was 242), 46 skipped, 0 changed
- [x] Scheduler tests specifically: 67 passed (was 66 — added the conformance test)
- [x] Manually verified conformance test fires on drift (removing an attribute from \`FakeRuntime\` causes the conformance test to fail with the missing name listed)

### DIFF
--- a/tests/scheduler/_fake_runtime.py
+++ b/tests/scheduler/_fake_runtime.py
@@ -1,0 +1,223 @@
+"""In-memory ``Runtime`` implementation for scheduler tests.
+
+The :class:`~kestrel.scheduler.scheduler.GenerationScheduler` is typed
+against the :class:`~kestrel.runtime.Runtime` protocol. ``FakeRuntime``
+gives the scheduler test suite a second consumer of that protocol so
+scheduler logic can be driven without instantiating ``MoondreamRuntime``
+(which requires real model weights, GPU resources, and vision/text
+encoders).
+
+Behaviour:
+- every Runtime method is implemented as a lightweight bookkeeping
+  call. Calls are recorded into public lists/dicts so tests can assert
+  ordering and arguments.
+- attributes default to small, plausible values (``max_batch_size=1``,
+  ``page_size=1``, etc.) but are mutable so individual tests can
+  override.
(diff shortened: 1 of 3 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

PR #50, opened 2026-05-05
Type the engine against the Runtime protocol

vikhyat:
## Summary

\`InferenceEngine\` still constructs \`MoondreamRuntime\` concretely (only the engine knows which runtime to instantiate today), but everywhere it *holds*, *returns*, or *accepts* a runtime reference is now typed against the \`Runtime\` protocol introduced in #48.

This is the engine-side counterpart to #48 (which did the scheduler) and lets us replace the engine's single-runtime hold-site with a multi-runtime registry in a follow-up without rewriting the type story.

## Changes

**\`kestrel/runtime/protocol.py\`** — adds two attributes the engine already reads on \`MoondreamRuntime\`:
- \`prefix_cache: Any\` — \`engine.py:234\` checks \`is not None\`
- \`graph_capture_lock: Any\` — \`engine.py:1274\` uses as a context manager

Both \`Any\` to avoid coupling the protocol to the \`RadixPrefixCache\` / \`threading.Lock\` concrete types — declaring them keeps the protocol an honest record of the surface a runtime must satisfy today.

**\`kestrel/engine.py\`** — annotation flips from \`MoondreamRuntime\` to \`Runtime\`:
- \`_AdmissionCoordinator.__init__\`'s \`runtime\` parameter
- \`InferenceEngine._runtime\` field
- \`InferenceEngine.runtime\` property return type
- \`_build_generation_request\`'s \`runtime\` parameter

The lone remaining \`MoondreamRuntime\` reference is the construction at \`engine.py:427\` — that's correct; the engine has to pick a concrete runtime to instantiate.

**\`tests/scheduler/_fake_runtime.py\`** — \`FakeRuntime\` gains the new attrs (\`prefix_cache=None\`, \`graph_capture_lock=contextlib.nullcontext()\`) so the conformance test in \`test_runtime_conformance.py\` still passes.

## Test plan
- [x] \`pyright -p . kestrel/engine.py kestrel/scheduler kestrel/runtime\` — 0 errors
- [x] \`pytest --ignore=tests/integration\` — 244 passed, 46 skipped, 0 changed
- [x] FakeRuntime conformance test (signature + attribute checks) still passes after adding the two new attrs

### DIFF
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -55,6 +55,7 @@ import torch
 from kestrel.config import RuntimeConfig
 from kestrel.device import set_device, synchronize
 from kestrel.moondream.runtime import MoondreamRuntime
+from kestrel.runtime import Runtime
 from kestrel.scheduler import (
     GenerationScheduler,
     GenerationRequest,
@@ -210,7 +211,7 @@ class _ReadyAdmission:
 class _AdmissionCoordinator:
     def __init__(
         self,
-        runtime: MoondreamRuntime,
+        runtime: Runtime,
         image_preprocessor: ImagePreprocessor,
         wake_event: threading.Event,
         fail_request: Callable[[_PendingRequest, BaseException], None],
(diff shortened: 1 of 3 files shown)


vikhyat:
@codex review

PR #51, opened 2026-05-05
Let multiple runtimes share a KVMemoryPool with a budget cap

vikhyat:
## Summary

Two changes that together let one process host more than one runtime behind a single KV memory budget:

1. **\`KVMemoryPool\` gains an optional \`budget_bytes\` parameter.** \`allocate_layer_kv\` raises \`MemoryError\` when an allocation would push the cumulative \`allocated_bytes\` past the cap. Existing unbudgeted construction stays a pure accounting layer with no enforcement, so current callers don't change behaviour.
2. **\`MoondreamRuntime\` accepts an optional \`kv_pool\` constructor kwarg.** When \`None\` it builds a private pool (current behaviour); otherwise it borrows the caller-supplied pool. A device mismatch raises immediately so wrong-pool wiring fails at construction rather than at first allocation.

Together these let a caller construct one pool with a total budget and pass it to multiple runtime instances. The pool counts every \`(k, v)\` allocation against the same cap, so neither runtime can over-allocate against the other.

## Implementation notes

- The pool now computes layer byte count from \`element_size(dtype)\` directly (no temporary tensor) so the precondition check happens before any allocation.
- A failed budget check leaves \`allocated_bytes\` unchanged so a caller can retry against a different pool / smaller cache without leaking state. New tests pin this.

## Test plan
- [x] \`pytest --ignore=tests/integration\` — 249 passed (was 244, +5 from \`tests/test_kv_pool.py\`), 46 skipped, 0 changed
- [x] New tests cover: unbudgeted accounting, multiple caches sharing one budget, budget exhaustion raises and doesn't advance counter, negative budget rejected, string device normalization
- [x] \`pyright -p . kestrel/kv_cache.py kestrel/moondream/runtime.py\` — 0 errors (2 unrelated pre-existing warnings)

### DIFF
--- a/kestrel/config.py
+++ b/kestrel/config.py
@@ -7,6 +7,8 @@ import re
 
 import torch
 
+from kestrel.device import resolve_device
+
 
 _SMALL_VRAM_THRESHOLD_BYTES = 24 * 1024**3
 _KV_CACHE_PAGES_SMALL_VRAM = 16384
@@ -142,11 +144,7 @@ class RuntimeConfig:
     def resolved_device(self) -> torch.device:
         """Return the torch device requested for inference."""
 
-        device = torch.device(self.device)
-        # Ensure CUDA devices have an explicit index for torch.cuda.set_device()
-        if device.type == "cuda" and device.index is None:
-            device = torch.device("cuda", torch.cuda.current_device())
-        return device
+        return resolve_device(self.device)
 
     def _validate_device_available(self) -> None:
         """Raise a clear error if the requested device isn't usable.
--- a/kestrel/device.py
+++ b/kestrel/device.py
@@ -18,6 +18,22 @@ from typing import Optional
 import torch
 
 
+def resolve_device(device: torch.device | str) -> torch.device:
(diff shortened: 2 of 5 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

Updated since the last review: `979c8c5` lifts the CUDA-index canonicalization out of `RuntimeConfig.resolved_device` into a shared `kestrel.device.resolve_device` helper, used from both. `KVMemoryPool(device='cuda')` now resolves to `cuda:N` and compares equal to the runtime's resolved device.

vikhyat:
@codex review

`c2013bb` addresses both P1s from the previous review:
- threading.Lock around budget check + reservation + counter update; race-tested with a barrier-synchronized two-thread test that asserts exactly one allocation wins and one raises when the pool is sized for one.
- weakref.finalize on each returned tensor so when it's collected the pool decrements; covers the partial-init-failure case where layers 0..N succeed and N+1 raises (GC of the prior caches releases their bytes). Also rolls back the reservation if torch.zeros itself raises mid-allocation.

PR #52, opened 2026-05-05
Scope prefix cache by runtime_id so multiple runtimes don't collide

vikhyat:
## Summary

When two runtimes share one prefix cache (e.g., a multi-runtime engine hosting different model variants behind a shared cache), entries from different models must not surface as hits for each other — token-id sequences are model-specific, and a cross-model hit would return KV pages with incompatible shape and meaning.

Add a \`runtime_id\` field to \`CacheNamespace\` so the radix cache trees are per-runtime by default for runtime-scoped callers. Default \`None\` keeps every existing caller working unchanged.

## Changes

- **\`CacheNamespace\`** gains a \`runtime_id: str | None = None\` field. Two namespaces with different \`runtime_id\` values resolve to separate trees inside \`RadixPrefixCache\`.
- **\`Runtime\` protocol** gains \`model_name: str\`. Stable identity for the runtime; \`MoondreamRuntime.model_name\` already returns the configured model variant. The multi-runtime engine routing that lands next will read the same field.
- All four \`CacheNamespace(...)\` construction sites in \`MoondreamRuntime\` now pass \`runtime_id=self.model_name\`.
- **\`FakeRuntime\`** exposes \`model_name\` (defaulting to \`"fake"\`) so the conformance test still passes.

## Tests

- New \`test_runtime_id_namespace\` in \`tests/prefix_cache/test_radix_cache.py\` — exercises scope isolation across two runtime ids and against the default \`runtime_id=None\` namespace.
- \`tests/test_page_table.py\` duplicate-prefill tests construct a bare \`MoondreamRuntime\` via \`__new__\` and exercise \`_finalize_cache_after_prefill\`. Updated the helper to seed a \`_cfg\` stub backing \`model_name\`, and migrated the manual \`cache.insert\` / \`match_prefix\` calls to use the same \`runtime_id\`-scoped namespace the finalize path now uses.

## Test plan
- [x] \`pytest --ignore=tests/integration\` — 252 passed (was 249, +3), 47 skipped, 0 changed
- [x] \`pyright -p . kestrel/prefix_cache kestrel/runtime kestrel/moondream/runtime.py\` — 0 errors
- [x] Verified scope isolation: two namespaces with different \`runtime_id\` insert/match independently; same-tokens-different-runtime returns no match

### DIFF
--- a/kestrel/moondream/runtime.py
+++ b/kestrel/moondream/runtime.py
@@ -786,6 +786,7 @@ class MoondreamRuntime:
         # Build namespace
         image_hash_int = int.from_bytes(image_hash[:16], "big")
         namespace = CacheNamespace(
+            runtime_id=self.model_name,
             lora_id=adapter_id,
             image_hash=image_hash_int,
         )
@@ -819,6 +820,7 @@ class MoondreamRuntime:
                 )
             cache_tokens = self._build_cache_tokens(tokens_list, image_hash, image_kv_length)
             namespace = CacheNamespace(
+                runtime_id=self.model_name,
                 lora_id=adapter_id,
                 image_hash=(
                     int.from_bytes(image_hash[:16], "big") if image_hash is not None else None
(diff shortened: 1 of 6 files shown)


vikhyat:
@codex review

PR #53, opened 2026-05-05
Let InferenceEngine host an externally-built runtime

vikhyat:
## Summary

The engine has been the only place that constructs \`MoondreamRuntime\`, which forced a 1:1 binding between engines and a single runtime class. This PR lets the caller build the runtime themselves and hand it to the engine, so **two engines (different model variants) can coexist in one process behind a shared \`KVMemoryPool\`** — each engine still owns its own scheduler thread and request queue, but the GPU's KV budget is common.

The expected wiring:

\`\`\`python
shared_pool = KVMemoryPool(device="cuda", budget_bytes=...)
md_runtime = MoondreamRuntime(md_cfg, kv_pool=shared_pool)
other_runtime = SomeOtherRuntime(other_cfg, kv_pool=shared_pool)

md_engine    = InferenceEngine(md_cfg,    runtime=md_runtime)
other_engine = InferenceEngine(other_cfg, runtime=other_runtime)
\`\`\`

## Changes

- \`InferenceEngine.__init__\` gains \`runtime: Runtime | None = None\`, plumbed through \`create()\`. When provided, \`_initialize\` skips the internal \`MoondreamRuntime\` construction.
- \`RuntimeConfig\` is still required (PhotonReporter telemetry uses it).
- Validates \`runtime.device == runtime_cfg.resolved_device()\` at construction so wrong-pool wiring fails immediately.
- Decouples the "already initialized" guard from the "runtime exists" check by introducing \`self._initialized\`. An injected runtime no longer short-circuits warmup or scheduler-thread setup.

## Test plan
- [x] \`pytest --ignore=tests/integration\` — 255 passed (was 252, +3), 47 skipped, 0 changed
- [x] New tests cover: injected runtime visible immediately on \`engine.runtime\`; device mismatch raises at construction; un-injected engine still raises on \`engine.runtime\` until \`_initialize\` runs
- [x] \`pyright -p . kestrel/engine.py\` — 0 errors

### DIFF
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -418,7 +418,6 @@ class InferenceEngine:
     async def _initialize(self) -> None:
         if self._initialized:
             return
-        self._initialized = True
         loop = asyncio.get_running_loop()
         self._loop = loop
         if self._runtime is None:
@@ -460,6 +459,10 @@ class InferenceEngine:
             )
             await self._photon_reporter.validate_api_key()
             self._photon_reporter.start()
+        # Flip the guard last so a failure partway through (runtime
+        # construction, warmup, photon) leaves the engine retry-able
+        # rather than wedged in a half-initialized state.
+        self._initialized = True
 
     async def _warmup_query_pipeline(self) -> None:
         """Ensure the high-level query path is exercised before serving traffic."""
@@ -986,7 +989,7 @@ class InferenceEngine:
         return future, req_id
 
     async def _ensure_started(self) -> None:
-        if self._runtime is None:
+        if not self._initialized:
             await self._initialize()
 
     async def _worker_loop(self) -> None:
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -326,14 +326,15 @@ class InferenceEngine:
         # An externally-built runtime is held from construction; otherwise
         # ``_initialize`` builds a MoondreamRuntime on first ``create``.
         self._runtime: Optional[Runtime] = runtime
-        # Two-phase init flags. ``_started`` flips once the scheduler
-        # thread + worker task are up so the warmup pipeline (which
-        # re-enters via ``query()`` → ``_ensure_started``) doesn't
-        # recurse back into ``_initialize``. ``_initialized`` flips at
-        # the very end so a partial failure (warmup, photon) can still
-        # be retried by another ``_initialize`` call.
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

`a25ecda` addresses both:
- `_ensure_started` now gates on `_initialized` (was `_runtime is None`), so an injected runtime still triggers worker task + scheduler thread + warmup.
- `_initialized = True` moved to the end of `_initialize`; a partial failure leaves the engine retry-able instead of wedged.

vikhyat:
@codex review

`592f293` splits init into two phases to break the warmup recursion:
- `_started` flips after the scheduler thread + worker task are up, before warmup. The warmup pipeline's recursive `_ensure_started` call sees `_started=True` and returns without re-entering `_initialize`.
- `_initialized` still flips at the very end (warmup + photon complete), so partial-init failures (e.g., warmup or photon-reporter throwing) leave the engine retry-able.

`_ensure_started` gates on `_started`; `_initialize`'s self-guard checks `_initialized`.

vikhyat:
@codex review

`c51d8f5` redesigns the init lifecycle:
- Drop `_started`. Use `_initializing` as an in-progress guard held inside try/finally.
- `_ensure_started` returns when `_initialized OR _initializing` is set, so the warmup pipeline's recursive call doesn't re-enter init *and* concurrent first-request callers don't both kick it off.
- Each step in `_initialize` is idempotent: runtime built only if None, scheduler thread only if not alive, worker task only if None or done, photon only if None.

A failure during warmup or photon validation now leaves the engine retry-able — the next `_ensure_started` triggers another `_initialize` that skips already-completed steps and re-runs the failed ones.

vikhyat:
@codex review

`6ca4e2c` addresses both P1s:
- PhotonReporter is built into a local, validated + started, then committed to `self._photon_reporter` only on success — a failed validate now leaves `_photon_reporter=None` and retries rebuild.
- Replaced `_initializing: bool` with `_init_task: asyncio.Task | None`. `_ensure_started` awaits the in-flight task via `asyncio.shield` when a different task has init in progress, so concurrent first-request callers actually wait. Warmup recursion is caught via `asyncio.current_task() is self._init_task` — same-task guard avoids deadlock without skipping the wait for cross-task callers.

PR #55, opened 2026-05-07
Add opt-in token logprobs to engine results

vikhyat:
## Summary
- Add private opt-in logprob support via `settings={"_logprobs": True}` and matching private submit APIs.
- Return `EngineResult.logprobs` / `SchedulerResult.logprobs` as `None` by default, or a generated-token-aligned `list[float]` when requested.
- Thread sampled-token logprobs through scheduler staging, D2H transfer, prefill/decode commits, and spatial coord/size decoding.
- Route all-greedy requests through the sampler when logprobs are requested, so greedy rows report model logprobs instead of `0.0`.
- Avoid passing `logprobs_out` unless requested so non-logprob traffic keeps the existing fast paths.

## Notes
- `_logprobs=True` requires a `kestrel-kernels` build containing the merged sampler logprob support from m87-labs/kestrel-proprietary#126.
- Coord/size token logprobs include the top-level coord/size token probability plus the selected spatial bin probability.
- Mixed scheduler batches may compute sampled logprobs for every row when any row opts in, but only opted-in requests expose values in results.

## Validation
- Local: `uv run pytest tests -q` -> `267 passed, 47 skipped`
- p2 focused kernel tests from m87-labs/kestrel-proprietary#126: `tests/test_sample_step_cuda.py` -> `20 passed, 1 warning`
- p2 sampler benchmarks showed no measurable regression for non-logprob calls because they continue using the non-logprob sampler entrypoint.

### DIFF
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -540,7 +540,7 @@ class InferenceEngine:
         image: Optional[np.ndarray | bytes] = None,
         temperature: Optional[float] = None,
         top_p: Optional[float] = None,
-        _return_logprobs: Optional[bool] = None,
+        _logprobs: Optional[bool] = None,
     ) -> EngineResult:
         future, _ = await self._submit_request(
             max_new_tokens=max_new_tokens,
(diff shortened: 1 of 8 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

PR #62, opened 2026-05-19
Pass compact MoE LoRA metadata

vikhyat:
Summary:
- Pass graph-stable MoE LoRA metadata through decode slots, `text_decoder`, `moe_mlp`, and `MoEModule`.
- Decode now has one LoRA metadata path for graph replay and non-graph execution: fixed-capacity buffers plus device-side `[active_count, active_rank]` metadata.
- Prefill uses the same batched MoE LoRA API with compact one-adapter metadata instead of the removed single-LoRA path.
- Add a runtime metadata unit test for compact route IDs and active adapter/rank metadata.

Depends on m87-labs/kestrel-proprietary#135.

Graph stability:
- CUDA graph replay keeps the same tensor addresses and launch topology.
- Per-step CPU metadata preparation writes compact route IDs and active metadata into existing slot buffers before replay.
- There is no separate eager-only compact path.

Performance status:
- Kernel-side measurements are in m87-labs/kestrel-proprietary#135.
- With corrected production rank semantics, the CuTe path is faster than the old Triton path on belka RTX 3090 for the measured production-shape cases.

Validation:
- belka RTX 3090: installed the updated local `kestrel-kernels` wheel into the Kestrel venv.
- `tests/moondream/test_moe.py`: 4 passed.


vikhyat:
@codex review

PR #65, opened 2026-05-22
Move query tokenization layout into per-model config

vikhyat:
## Problem

The query skill encoded per-model prompt tokenization as `model_name == "moondream2"` branches. MD2 was trained with a doubled `answer_id` at the question boundary (an HF tokenizer-pipeline artifact); `build_prompt_tokens` reproduced it with an `is_md2` branch, and `consume_step` force-fed the suffix after reasoning for *every* model. Adding a new model meant editing the skill — and knowing the artifact existed.

## Change

`templates["query"]` now declares the token layout as explicit slots, and the skill renders them with no model branches:

- `prefix` — before the question
- `answer_prefix` — question → answer (non-reasoning)
- `reasoning_prefix` — question → reasoning
- `post_reasoning_prefix` — injected after the generated `answer_id`, before the answer

MD2 and MD3 each carry their own `query` template. The MD2 artifact is now visible and commented in `DEFAULT_MOONDREAM2_CONFIG`, where it belongs, instead of hidden in a string compare in the skill. A new model ships a `query` block and touches neither `query.py` nor `config.py`.

## Behavior

Token-identical for MD2 across both modes — the new slot values reproduce the previous assembly exactly (checked against all four model × mode cases). MD3 stops receiving the post-reasoning `answer_id` injection: it was redundant (MD3 generates `answer_id` itself to end reasoning) and MD3 never benefited from it.

ChartQA test split, CoT, evee2 / B200:

| | MD2 | MD3 |
|---|---|---|
| before | 77.80% | 85.92% |
| after | 77.80% | 85.88% |

MD2 unchanged; the MD3 delta is the post-reasoning injection removal — 1 question of 2500, within run-to-run noise.

## Out of scope

`suppressed_token_ids` still carries a `model_name == "moondream2"` branch — the same pattern, with the same config-driven fix shape (per-phase suppressed-token lists). Left as a separate change.

### DIFF
--- a/kestrel/moondream/config.py
+++ b/kestrel/moondream/config.py
@@ -51,7 +51,12 @@ class TokenizerConfig:
                 "normal": [1, 32708, 2, 6382, 3],
                 "long": [1, 32708, 2, 4059, 3],
             },
-            "query": {"prefix": [1, 15381, 2], "suffix": [3]},
+            "query": {
+                "prefix": [1, 15381, 2],
+                "answer_prefix": [3],
+                "reasoning_prefix": [4],
+                "post_reasoning_prefix": [],
+            },
             "detect": {"prefix": [1, 7235, 476, 2], "suffix": [3]},
             "point": {"prefix": [1, 2581, 2], "suffix": [3]},
             "segment": {"prefix": [1, 17374, 2], "suffix": [3]},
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

PR #69, opened 2026-05-25
Dispatch runtime construction through ModelSpec.runtime

vikhyat:
## Summary

- Adds a ``runtime`` field to ``ModelSpec`` holding the constructor the engine invokes as ``runtime(cfg, **kwargs)`` to produce a concrete ``Runtime`` for the given model. Both Moondream specs set it to ``MoondreamRuntime``.
- Replaces the hardcoded ``MoondreamRuntime(self._runtime_cfg, max_lora_rank=...)`` in ``InferenceEngine._initialize`` with ``get_spec(cfg.model).runtime(cfg, max_lora_rank=...)``, and drops the now-unused top-level ``MoondreamRuntime`` import from ``engine.py``.
- Lets external model packages plug in by registering a spec whose ``runtime`` points at their own class — no engine changes required to add a new model family.

The engine still imports other Moondream-specific helpers (image preprocessing, token types, LoRA adapter); decoupling those is a separate, larger follow-up.

## Verification

- ``pytest tests/`` on evee2 (B200): 316 passed, 4 skipped. The pre-existing ``tests/test_moe_lora.py`` kernel-skew failure is unchanged.
- Confirmed ``get_spec("moondream2").runtime`` and ``get_spec("moondream3-preview").runtime`` both resolve to ``MoondreamRuntime``.

### DIFF
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -55,7 +55,6 @@ import torch
 from kestrel_kernels import get_runtime
 from kestrel.config import RuntimeConfig
 from kestrel.device import set_device, synchronize
-from kestrel.models.moondream.runtime import MoondreamRuntime
 from kestrel.runtime import Runtime
 from kestrel.scheduler import (
     GeneratedPrefix,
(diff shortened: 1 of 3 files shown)


vikhyat:
@codex review

PR #73, opened 2026-06-01
Split Runtime protocol by execution shape

vikhyat:
## Why

The engine is moving toward hosting multiple co-resident models in one process. Today's `Runtime` protocol is implicitly autoregressive — it bakes in prefill slots, a KV/prefix cache, and a per-token decode loop — even though its name is generic. To host models with a different execution shape (a single forward, no decode loop), the contract needs to separate "what every runtime is" from "what a prefill/decode runtime is."

## What

Introduce a three-layer protocol stack in `kestrel/runtime/protocol.py`:

- **`Runtime`** — the minimal base every runtime satisfies: `model_name`, `device`, and an `execution_shape` tag.
- **`AutoregressiveRuntime(Runtime)`** — the existing prefill/decode surface, moved verbatim under the base (KV cache, slots, prefix cache, `classify_prefill` / `prepare_sequence` / `launch_prepared_batch` / `decode_with_slot`, …). This is what `GenerationScheduler` consumes.
- **`SinglePassRuntime(Runtime)`** — runtimes that fulfill a request with one forward: `run(task, inputs)` plus image preprocessing.

A new `ExecutionShape` enum (`AUTOREGRESSIVE` / `SINGLE_PASS`) lets the engine route by shape via a lookup rather than `isinstance` checks.

The AR-specific `Runtime` type references in the engine, scheduler, and request types are renamed to `AutoregressiveRuntime`. `MoondreamRuntime` declares `execution_shape = AUTOREGRESSIVE`. The runtime-conformance check walks the protocol MRO so members inherited from the base stay covered.

No behavior change — protocol split plus mechanical rename. Existing scheduler and engine tests pass unchanged (`tests/scheduler/` — 93 passed).

### DIFF
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -56,7 +56,7 @@ import torch
 from kestrel_kernels import get_runtime
 from kestrel.config import RuntimeConfig
 from kestrel.device import set_device, synchronize
-from kestrel.runtime import Runtime
+from kestrel.runtime import AutoregressiveRuntime
 from kestrel.scheduler import (
     GeneratedPrefix,
     GenerationScheduler,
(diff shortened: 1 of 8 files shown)


vikhyat:
@codex review

PR #75, opened 2026-06-02
Service single-pass jobs on the owner thread

vikhyat:
## Why

The engine drives one model family today: autoregressive prefill/decode. The next execution shape — single-pass models that fulfill a request with one forward and no decode loop — needs to run on the **same owner thread** that owns all CUDA state, without a second scheduler and without overlapping autoregressive GPU work. This PR adds that path. It's the only PR in the series that touches the scheduler loop, kept isolated for that reason.

## What

- **`_SinglePassJob`** + a thread-safe **`_single_pass_queue`** on the engine — separate from the AR request queue (no admission, KV reservation, or prefill batching).
- **`_submit_single_pass(model, task, inputs)`** — validates the model is registered and single-pass (`execution_shape`), enqueues a job, wakes the loop, and awaits its future. This is the internal seam the multi-model surface will call; there is no public API yet.
- **`_scheduler_loop`** services **one** single-pass job per iteration at a safe point — only when no AR forward is in flight (`pipeline.has_launch_in_flight()` is false) — so the single forward never overlaps autoregressive GPU work and all CUDA stays on one thread (the single-owner invariant). `should_exit` waits for the queue to drain; the `finally` fails any leftover jobs on shutdown.

## Tests

CPU end-to-end tests drive the **real** `_scheduler_loop` with a `FakeRuntime` as the autoregressive default plus a stub single-pass runtime registered alongside it: a job round-trips through the loop, a failing job propagates its exception, and submission rejects unknown / wrong-shape models. Full engine + scheduler suite green (119 passed).

### DIFF
--- a/kestrel/engine.py
+++ b/kestrel/engine.py
@@ -56,7 +56,7 @@ import torch
 from kestrel_kernels import get_runtime
 from kestrel.config import RuntimeConfig
 from kestrel.device import set_device, synchronize
-from kestrel.runtime import AutoregressiveRuntime
+from kestrel.runtime import AutoregressiveRuntime, ExecutionShape
 from kestrel.scheduler import (
     GeneratedPrefix,
     GenerationScheduler,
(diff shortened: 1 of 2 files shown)


vikhyat:
Closing in favor of the executor/kernel paradigm (design updated in kestrel-proprietary#176).

This PR serviced single-pass work **synchronously** between AR decode steps: the owner thread pulled a job, called `runtime.run()`, and blocked on that forward + its D2H before resuming AR. That stalls AR decode for the duration of every single-pass forward — no interleaving, no slots, no concurrency. It's the wrong paradigm, and building on it would mean rewriting this area once a real single-pass model (the segmentation runtime) lands.

The replacement establishes the final-state model up front so nothing switches paradigms mid-stream:
- Engine = **kernel**, model/runtime = **driver**; the per-shape machinery is an **executor** (a kernel-side lane wrapping one driver).
- Uniform value-returning `Executor` protocol (`submit` / `advance() -> TickResult` / `shutdown`); the kernel loop is a fold over executors that performs delivery effects.
- The driver's `forward()` **enqueues kernels without a host sync**; the executor owns D2H + completion + slots, so single-pass forwards **interleave** with AR decode on the shared stream (async-launch + deferred-collect — the AR pipeline's own trick, generalized across lanes).
- AR gets wrapped as an executor with its pipeline internals untouched.

Next PR rebuilds this as the executor + kernel-loop foundation (AR wrapped, no behavior change), then a follow-up adds the single-pass executor with real async launch/collect.

PR #79, opened 2026-06-02
Move input validation into model-owned skills

vikhyat:
A skill describes one of a model's capabilities, so it belongs with the model rather than in the model-agnostic kernel. Today the kernel constructed and validated each model's request types directly; this moves that ownership to the model and leaves the kernel a clean, model-independent contract.

## What changed

- **Relocate concrete skills into the model package.** The `query`/`caption`/`detect`/`point`/`segment` modules move out of the top-level `kestrel.skills` package and under the model. The top-level package now holds only the model-agnostic contract: `SkillSpec`, `SkillState`, `SkillRegistry`, the value types skills exchange with the kernel (`DecodeStep`, `BuiltRequest`, `SkillFinalizeResult`), and the autoregressive sampling defaults.

- **Validation/request-building moves into the skill.** Each skill implements `build_request`, which validates the raw inputs and assembles the model's own request type plus sampling params. The engine verbs become thin adapters that pack their arguments into a `prompt` dict and delegate — the kernel no longer imports, constructs, or inspects any model-specific request type.

- **The skill registry is static model metadata on `ModelSpec`** (symmetric with the existing runtime factory). Because capabilities are declared as metadata, the engine can resolve them *without* building the runtime: input validation and `tasks`/`supports` introspection now work before startup, and invalid input fails fast without loading model weights.

- **Dead code removed along the way:** the scheduler's model-specific skill fallback (the kernel held a hard-coded default skill), the unused per-skill `*Settings` dataclasses, and an unused CLI module.

## Notes

`InferenceEngine`'s public verb signatures and the `ModelHandle` surface are unchanged; this is an internal reshaping of where validation lives. A drift guard test pins the handle verbs to the engine verbs.


vikhyat:
@codex review

vikhyat:
@codex review

PR #81, opened 2026-06-03
Build co-hosted models at startup; generalize ModelSpec for single-pass

vikhyat:
Wires up multi-model construction so registered single-pass models are actually built and served — the prerequisite for dropping in a single-pass model. Previously the engine built only the *default* model's runtime; non-default single-pass runtimes were never constructed from a spec (only injected directly in tests), so the single-pass executor/lane machinery (#77) was never exercised in production.

## What changed

- **`Engine.create(cfg, models=[...])`** builds a runtime for every configured model at startup — the default autoregressive model plus co-hosted models. The scheduler already creates one single-pass lane per single-pass runtime, so the lanes follow automatically. Omitting `models=` is fully back-compat (today's single-model behavior). The configured set is order-preserving (default first) and deduplicated; `engine.model(id)` works for a co-hosted id before startup.
- **`ModelSpec` generalized** (the doc's step 8): only `name`, `runtime`, and `skills` are required now; the autoregressive/HuggingFace bootstrap fields (`repo_id`, `filename`, `checkpoint_format`, `tokenizer_id`, `default_config`) are optional. The kernel never reads them — they're consumed by a specific runtime family's loader.
- **`ensure_model_weights` returns `None`** when a spec declares no weight file, so a single-pass spec bypasses the AR checkpoint downloader and its runtime factory owns loading (e.g. `snapshot_download`). Each co-hosted model is built with a per-model config (its own `model` id; `model_path` reset so it resolves for that model). `max_lora_rank` is forwarded only to the default AR model (single-pass factories don't accept it).

## Validation

- Unit tests (`tests/test_multi_model.py`): `models=` tracking/dedup, construction of every configured model from specs incl. a tokenizer-free single-pass spec, per-model config, `max_lora_rank` routing, and idempotent skip of already-built runtimes.
- Full suite green (the 3 `test_moe_lora` failures are a pre-existing kernel-build skew, unrelated to this change).
- GPU end-to-end: a real default autoregressive lane and a co-hosted single-pass lane interleave correctly (`query` → `run` → `query`), and back-compat (`models` omitted) still builds and serves the default.

`InferenceEngine`'s public surface is unchanged except the new optional `models=` argument.


vikhyat:
@codex review

PR #83, opened 2026-06-04
Reorder decode tick to commit -> finalize -> launch

vikhyat:
## What

In `GenerationScheduler.advance()`, the decode tick runs three blocks. This reorders them from **launch -> commit -> finalize** to **commit -> finalize -> launch**, moving the next-forward launch to the end. Every bit of existing logic is preserved (prefill-launch handling, the `can_launch` / `free_slot_id` guards, the defensive `AssertionError`, the `not progressed` stall check); only the order of the three blocks changes, plus the docstring/comments.

## Why

Launching the next forward *last* enqueues it behind this step's sampling, so it runs on the GPU while the **next** tick performs its commit + finalize. This overlaps the commit and launch CPU work with the forward instead of paying the launch on the critical path between forwards.

## Invariants preserved

- **commit-before-finalize**: the commit advances skill state (`consume_step`), so the forward we finalize samples under the correct constrained-decoding mask. The new order keeps commit before finalize.
- **depth-1 speculation (zombies)**: the next forward is launched only after committing the step two behind it, so a sequence that ends still rides exactly one in-flight step — no extra zombie waste.

## Measured

On a B200 (moondream2, interleaved A/B): ~+1% aggregate decode throughput at high concurrency (32 streams), neutral at concurrency 1, output-identical. Output equivalence was verified in prior work — the reordered pipelined loop produces byte-identical outputs vs the depth-1 blocking loop across query/caption/point/detect and a 6-way concurrent batch with varying lengths.

### DIFF
--- a/kestrel/scheduler/scheduler.py
+++ b/kestrel/scheduler/scheduler.py
@@ -323,14 +323,24 @@ class GenerationScheduler:
         sequences admitted). Callers can keep invoking ``advance`` while it
         returns ``True`` to drain ready work before sleeping.
 
-        Phase 1 ordering (design doc §5):
-        1. Launch forward (if none in-flight) - forward doesn't need mask
-        2. Commit previous step (updates skill state via consume_step)
-        3. Finalize sampling (compute mask with updated skill state)
+        Decode-tick ordering: commit -> finalize -> launch.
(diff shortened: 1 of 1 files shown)


vikhyat:
@codex review

PR #84, opened 2026-06-04
Stage the constrained-decode sampling mask on the copy stream

vikhyat:
## Summary

The per-step constrained-decode sampling mask was applied **synchronously on the compute stream**. For each constrained batch row, `_sample_batch` built a Python list of the row's allowed (and suppressed) token ids and did `torch.tensor(<python list>, device=cuda)` — a pageable H2D copy that blocks the CPU until it lands. That cost recurred on **every decode step**, up to once per batch row, so the CPU stall scaled with batch size and consumed a growing fraction of each (shrinking) forward as the GPU got faster.

This moves the mask off the compute stream's critical path:

- Right after `commit_step` (so `skill_state` is current) and **while the forward is in flight**, `_build_mask` fills a per-row boolean *disallow* mask over the vocabulary (`True` => force that token's logit to `-inf`) into the slot's pinned host buffer and uploads it **async on the slot's copy stream**, recording `mask_ready_event`.
- Sampling then only `wait_event`s on that event and applies a single vectorized `logits.masked_fill_(disallow, -inf)` — no per-row `torch.tensor`, no per-row apply loop, no synchronous H2D.
- When nothing constrains the step, the mask build and upload are skipped entirely (`disallow`/`event` are `None`).

The mask is carried as a small `_MaskPlan` from the advance loop into `finalize_sampling` / `_finalize_sampling_on_stream` / `_sample_batch`. Prefill (once per request, off the per-token critical path) keeps building and applying its mask inline.

Output is unchanged: token-hash equivalence verified across `query`/`caption`/`point`/`detect` plus a 6-way concurrent batch — byte-identical to the prior path.

## Throughput

Aggregate decode-throughput wins by concurrency (c):

| GPU | c=1 | c=8 | c=16 | c=32 |
| --- | --- | --- | --- | --- |
| RTX 3090 | +5.4% | +6.3% | +7.4% | +11.3% |
| B200 | +16.3% | +20.7% | +24.0% | +33.7% |

The win grows with batch size (the old path did up to one synchronous H2D per row) and with GPU speed (a fixed CPU stall is a larger fraction of a shorter forward).

### DIFF
--- a/kestrel/models/moondream/decode_slot.py
+++ b/kestrel/models/moondream/decode_slot.py
@@ -144,6 +144,15 @@ class DecodeSlot:
     step_done_event: torch.cuda.Event = None  # type: ignore[assignment]
     commit_done_event: torch.cuda.Event = None  # type: ignore[assignment]
 
+    # Constrained-decode mask: a per-row boolean "disallow" mask over the
+    # vocabulary (True => force this token's logit to -inf). Built from
+    # skill_state right after commit and uploaded async on the copy stream
+    # during the forward; sampling waits on ``mask_ready_event`` then applies it
+    # with a single masked_fill_, so the per-step mask H2D never blocks the
+    # compute stream and there is no per-row apply loop.
+    disallow_mask: CpuGpuBuffer = None  # type: ignore[assignment]
+    mask_ready_event: torch.cuda.Event = None  # type: ignore[assignment]
+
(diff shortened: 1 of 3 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

PR #88, opened 2026-06-06
Start engine before handle task/shape checks for co-hosted models

vikhyat:
Addresses the Codex P2 from #81 (whose work is already in `main` as `231ca12`, so this is a small follow-up PR).

## The bug

`ModelHandle.run()` and the capability verbs inspected `_runtimes` / `tasks` **before** triggering startup. With eager `models=[...]` construction, a co-hosted model's runtime only exists after startup, so a pre-start handle call on one:

- `engine.model("sp").run(task, ...)` → `_require(task)` → `_tasks_for("sp")` → runtime not built → `ValueError("Unknown model 'sp'")`
- `engine.model("sp").segment(...)` → `_capability` sees `runtime=None` → falls to the autoregressive path → `NotImplementedError("not yet routable")`

Both violate the handle's documented contract ("a handle can be taken before the engine is started; the handle's async methods trigger startup"). `engine.run("sp", ...)` worked only because it calls `_ensure_started()` first.

## The fix

`run()` and `_capability()` now `await self._engine._ensure_started()` before checking tasks/shape, so the co-hosted runtime is built and its tasks/`execution_shape` are known. No change for the default model (already started lazily via the skill path) or for already-started engines.

## Validation

- New regression tests: pre-start `run()` and a pre-start capability verb on a co-hosted single-pass model both trigger startup and dispatch (rather than raising). 
- GPU end-to-end: on a fresh, unstarted engine with `models=["<default>", "<stub-sp>"]`, `engine.model("stub-sp").run(...)` and `.segment(...)` both build the runtimes and dispatch to the single-pass lane.
- Full suite green (the 3 `test_moe_lora` failures are a pre-existing kernel-build skew, unrelated).

### DIFF
--- a/kestrel/engine/handle.py
+++ b/kestrel/engine/handle.py
@@ -85,7 +85,12 @@ class ModelHandle:
         verbs dispatch to for that shape. Autoregressive models go through
         the skill path instead — ``engine.run`` would reject them, so reject
         here with a clearer message.
+
+        Starts the engine first: a co-hosted model's runtime (and thus its
+        tasks/shape) only exists once built, so the task check below must run
+        against a started engine, not raise "unknown model" pre-start.
         """
+        await self._engine._ensure_started()
         self._require(task)
         runtime = self._engine._runtimes.get(self._model)
         if runtime is not None and runtime.execution_shape is not (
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

PR #101, opened 2026-06-12
Fix profiler scripts after runtime/decoder/vision API drift

vikhyat:
The standalone profilers had drifted from the current APIs and no longer ran:

- `scripts/profile_decode.py`
  - `MoondreamRuntime` now requires `kv_pool`; construct a `KVMemoryPool` for it.
  - The NVTX-instrumented decoder mirror now absorbs unknown kwargs (so future decoder kwarg additions don't break loading) and forwards `moe_lora_metadata` to `moe_mlp`, matching the current signature — `lora_slot_ids`/`single_lora_id` are no longer MoE args.
- `scripts/profile_vision_encoder.py`
  - Follow the `_layernorm_bias_into` rename.
  - Print a steady-state CUDA-event `PROFILE: vision_forward avg_ms=` line after warmup, so the script gives a usable end-to-end number without wrapping it in nsys.

Verified on a B200 (evee1) against Moondream 3: decode profiling at bs1/16/64 and the vision profile both run end to end.

### DIFF
--- a/scripts/profile_decode.py
+++ b/scripts/profile_decode.py
@@ -116,6 +116,9 @@ def patch_text_decoder_with_nvtx():
         lora_workspace=None, lora_slot_ids=None, single_lora_id=None,
         dense_lora_scratch=None,
         scratch_pool=None,
+        # Absorb decoder kwargs added after this mirror was written so the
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

PR #104, opened 2026-06-15
Allow strided paged KV update inputs

vikhyat:
## Summary

- accepts BSHD paged-KV update tensors with compact head/head_dim layout and arbitrary token stride
- flattens update inputs with `as_strided` instead of requiring `.view()`-compatible contiguous storage
- covers fused-QKV value slices and batched decode value slices in `PagedKVCache.update` tests

## Why

Qwen's fused QKV projection produces value slices whose token stride includes the full fused projection width. The KV write kernel can handle that layout as long as each token's `[heads, head_dim]` tile is compact, so the engine should not force model runtimes to materialize a contiguous copy before writing KV.

## Validation

- `p1`: `tests/test_kv_pool.py::test_paged_cache_update_accepts_fused_projection_value_slice`
- `p1`: `tests/test_kv_pool.py::test_paged_cache_update_accepts_batched_decode_value_slice`

### DIFF
--- a/kestrel/kv_cache.py
+++ b/kestrel/kv_cache.py
@@ -27,6 +27,33 @@ def _cdiv(x: int | float | torch.Tensor, multiple: int | float | torch.Tensor):
     return (x + multiple - 1) // multiple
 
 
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

PR #107, opened 2026-06-17
Drop removed fallback_to_torch kwarg from Moondream layernorm calls

vikhyat:
## What

kestrel-kernels removed the `fallback_to_torch` parameter from `layernorm_bias` / `layernorm_bias_into` — the kernel now handles the supported configs directly (2D/3D, bf16, the SigLIP `N=1152` vision width via the `reload_x` variant) and raises on anything unsupported. The Moondream **text** (`models/moondream/layers.py`) and **vision** (`models/moondream/vision.py`) paths still passed `fallback_to_torch=True`.

Against current kestrel-kernels this makes every Moondream `layer_norm` raise:

```
TypeError: layernorm_bias() got an unexpected keyword argument 'fallback_to_torch'
```

which crashes `moondream3-preview` at decode CUDA-graph capture (and would fail eager decode too). Qwen is unaffected (RMSNorm, different kernel).

## Change

- Drop the stale `fallback_to_torch=True` from both call sites.
- Update the two comments that claimed a torch fallback (no longer true).

## Verified

With this applied to a current kestrel-latest + kestrel-kernels env on H100, `moondream3-preview` ChartQA runs cleanly again (smoke 64: Human acc 64.06%, 41.6 req/s; full run in progress). No other `fallback_to_torch` references remain in the tree.

### DIFF
--- a/kestrel/models/moondream/layers.py
+++ b/kestrel/models/moondream/layers.py
@@ -38,10 +38,10 @@ class LayerNormWeights:
 
 
 def layer_norm(x: torch.Tensor, w: LayerNormWeights) -> torch.Tensor:
-    # Cross-arch: dispatches to the .so kernel on CUDA, the Metal
-    # kernel on MPS. Anything else (or any unsupported config — fp16 on
-    # CUDA, non-mul-of-8 last dim) falls through to ``F.layer_norm``.
-    return _layernorm_bias(x, w.weight, w.bias, fallback_to_torch=True)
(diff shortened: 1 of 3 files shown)


vikhyat:
@codex review

vikhyat:
Good catch — fixed in 35de56c. `scripts/profile_vision_encoder.py` was missed in the #101 drift-fix pass (it's an unreferenced but maintained manual profiler, not dead code). Dropped the kwarg there too; a repo-wide grep now shows zero `fallback_to_torch` references. @codex review

PR #108, opened 2026-06-17
Capture Moondream paged-KV metadata in the decode CUDA graph

vikhyat:
## What

Moves `populate_paged_kv_metadata` out of the live `_prepare_decode_graph_step` and into `_run_decode_forward`, which is captured by the decode CUDA graph. This was the only per-step paged-KV metadata op still launched eagerly each decode step — `build_slot_mapping` right next to it is already built inside the captured forward, so this just brings the metadata build in line. Only the host-side LoRA metadata stays live in `_prepare_decode_graph_step`.

## Why it's safe

`populate_paged_kv_metadata` reads the static (pre-reserved at prefill) page table and the engine-written `batch_idx`/`input_pos` buffers, and writes the slot's fixed metadata buffers. A graph replay reproduces it identically — same argument that already justifies capturing `build_slot_mapping`.

## Measurement (honest: neutral)

ChartQA is heavily prefill-bound on `moondream3-preview` (~768 input vs ~3.3 output tokens/request), so per-step decode prep is a tiny slice. 2×2 interleaved A/B on H100 (pristine `main` vs this change, both with the `fallback_to_torch` layernorm fix; `--max-batch-size 4 --disable-prefix-cache`):

| | Accuracy | Decode tok/s | req/s |
|---|---|---|---|
| baseline ×2 | 84.08% / 84.12% (2102 / 2103) | 132.25 / 131.81 | 39.88 / 39.74 |
| this change ×2 | 84.04% / 84.12% (2101 / 2103) | 131.76 / 132.36 | 39.73 / 39.94 |

Accuracy is within the baseline's own run-to-run band (2102↔2103) and throughput is neutral. The value is consolidating the per-step launch into the single graph replay and matching the Qwen decode runtime — it should help decode-bound serving (long generation), which ChartQA does not exercise.

Depends on (or pairs with) #107, which is required for Moondream to run against current kestrel-kernels.

### DIFF
--- a/kestrel/models/moondream/runtime.py
+++ b/kestrel/models/moondream/runtime.py
@@ -2087,17 +2087,10 @@ class MoondreamRuntime:
         slot.meta.lora_slot_ids.cpu[batch_size:graph_batch_size].zero_()
 
     def _prepare_decode_graph_step(self, slot: DecodeSlot, batch_size: int) -> None:
+        # Only host-side LoRA metadata stays live here; the paged-KV metadata is
+        # built inside the captured decode graph (see _run_decode_forward).
         if self._lora_workspace is not None:
             self._prepare_decode_lora_metadata(slot, batch_size)
-
-        # Build per-step paged-KV metadata buffers.
-        batch_idx = slot.meta.batch_idx.gpu[:batch_size]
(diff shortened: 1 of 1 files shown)


vikhyat:
@codex review

PR #109, opened 2026-06-17
Add PackedBuffer: stage many mixed-dtype fields in one H2D copy

vikhyat:
## What

`PackedBuffer` packs several named tensors of **mixed dtype and shape** into a single `uint8` CPU/GPU buffer pair, so a group of small metadata tensors stages to the device in **one `cudaMemcpyAsync`** instead of one copy per field. Each field is an aligned `view(dtype).view(shape)` slice exposing the same `.cpu`/`.gpu`/`.np` surface as `CpuGpuBuffer`, so existing fill code (`field.np[...] = ...` on the host, `field.gpu[...]` on the device) works unchanged — only the copy collapses to a single `copy_to_gpu()` / `copy_to_cpu()`.

A single byte-backed store means there's no need to group fields by dtype; int64/int32/float/etc. all live in one buffer and ship together.

## Why

Profiling Qwen ChartQA showed it's CPU-launch-bound. The H2D metadata copies are one source of per-launch overhead. This is the reusable primitive for collapsing them, intended for per-step / per-batch staging across **prefill and decode, H2D and D2H, for both Qwen and Moondream**. First adopter is Qwen prefill (separate PR, ~11 copies → 1).

Note on expectations: on a fast, well-pipelined device (H100) the launch reduction is largely hidden by GPU compute, so it reads as neutral throughput there. The win is expected on host-bound / weaker-CPU targets (Thor, Orin, Mac MPS, low-core hosts) where per-launch CPU cost is on the critical path.

## Tests

`tests/test_packed_buffer.py`: shared-storage host writes, disjoint mixed-dtype regions, single-copy round-trip, bf16 (no numpy view), unknown-field error, and a CUDA round-trip (skipped without a GPU). All pass on CPU.

### DIFF
--- a/kestrel/utils/__init__.py
+++ b/kestrel/utils/__init__.py
@@ -1,5 +1,5 @@
 """Utility helpers for Kestrel."""
 
-from .buffers import CpuGpuBuffer
+from .buffers import CpuGpuBuffer, PackedBuffer, PackedField
 
-__all__ = ["CpuGpuBuffer"]
+__all__ = ["CpuGpuBuffer", "PackedBuffer", "PackedField"]
--- a/kestrel/utils/buffers.py
+++ b/kestrel/utils/buffers.py
@@ -1,6 +1,9 @@
 """Shared host/device buffer helpers."""
 
+from __future__ import annotations
 
+from collections.abc import Iterable, Sequence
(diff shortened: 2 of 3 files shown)


vikhyat:
@codex review

PR #112, opened 2026-06-22
Add shared ChatSkill (OpenAI-style multi-turn chat)

vikhyat:
## Summary
Adds a shared, model-agnostic `ChatSkill` exposing a multi-turn, OpenAI-style chat interface (`handle.chat(messages=[...])`). A model opts in by exposing a `chat()` prompt template; the output is shaped as an OpenAI chat-completion *choice* so a future `/v1/chat/completions` route is a thin wrapper. No HTTP server is added here — this is the client-side capability.

## What's included
- `ChatSkill` / `ChatSkillState` (`kestrel/skills/chat.py`): plain-Python message validation, single-image extraction from `image_url` content parts, prompt-token rendering, reasoning/answer decode, streaming deltas, and finalize to `{"message": {...}, "finish_reason": ...}`.
- `ChatTemplate` + `chat()` on the `PromptTemplate` protocol. Models with a native multi-turn (ChatML-style) format render natively; Moondream — no trained chat format — flattens the conversation into the single-turn `query` prompt.
- `BuiltRequest.image` so a skill can surface media pulled from its own prompt; threaded through `_run_skill`.
- `SkillState.stop_token_ids()`, consulted by the scheduler, so a turn can stop on a turn-end token distinct from `eos_id`.
- `chat` verb on `ModelHandle` and the engine; `ChatSkill` registered for Moondream.

## Validated
New unit tests (`tests/test_chat_skill.py`) plus the existing skill/handle/registry/scheduler/engine suites pass. Live multi-turn chat verified on Moondream (text + image via content parts) and a ChatML-format model (multi-turn memory; turn-end stop).

### DIFF
--- a/kestrel/engine/_types.py
+++ b/kestrel/engine/_types.py
@@ -301,10 +301,17 @@ class _ReadyAdmission:
     prefix_cache_hit: bool
 
 
-def _hash_image(image: np.ndarray | bytes) -> bytes:
-    """SHA-256 over the raw image input for prefix-cache keying."""
-    raw = image.tobytes() if isinstance(image, np.ndarray) else image
-    return hashlib.sha256(raw).digest()
+def _hash_image(image: "np.ndarray | bytes | Sequence[np.ndarray | bytes]") -> bytes:
+    """SHA-256 over the raw image input(s) for prefix-cache keying.
+
+    A chat request may carry several images (an ordered list); hash them in
+    order so the key reflects the full set.
+    """
+    items = image if isinstance(image, (list, tuple)) else [image]
+    digest = hashlib.sha256()
+    for one in items:
+        digest.update(one.tobytes() if isinstance(one, np.ndarray) else one)
+    return digest.digest()
 
 
 class EngineRequest(Protocol):
--- a/kestrel/engine/core.py
+++ b/kestrel/engine/core.py
@@ -1693,8 +1693,11 @@ class InferenceEngine:
         if req.image is None and image_crops is None:
             image_length = 0
         elif isinstance(req.image, (list, tuple)):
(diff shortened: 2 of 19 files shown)


vikhyat:
@codex review

vikhyat:
Addressed the P2 (suppression parity on the flattened chat path) in 065550e. Rather than the kernel branching on model specifics, Moondream now customizes chat via a subclass: `MoondreamChatSkill` / `MoondreamChatSkillState` flatten into the query prompt and reuse `QuerySkillState`, so the HF-parity `suppressed_token_ids` masking (and grounded reasoning) is inherited rather than reimplemented. The base `ChatSkill` now handles only models with a native `chat()` template. Added a test asserting the inherited md2 masking.

vikhyat:
@codex review

vikhyat:
Fixed the image_url path issue in dcdf7af. `_coerce_image` no longer reads server filesystem paths or guesses bare base64 from message content — `image_url.url` / `image` parts now accept only data: URLs and in-memory bytes/ndarray (a caller with a file loads it themselves). Added a test asserting a filesystem path in `image_url.url` is rejected.

vikhyat:
Tightened further in 682da55: dropped HTTP(S) URL handling too. A string image must now be a base64 `data:` URL — no remote fetches, no filesystem reads. In-memory bytes/ndarray still pass through for Python callers. Test now asserts both filesystem paths and http URLs are rejected.

vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

PR #116, opened 2026-06-29
Improve Photon CUDA diagnostics

vikhyat:
## Summary

- Raise Kestrel's minimum supported PyTorch version to 2.8 so the default CUDA 12 PyTorch runtime has the CUDA Library API used by kestrel-kernels.
- Improve CUDA-device validation when PyTorch was built with CUDA but cannot initialize it.
- Include the PyTorch CUDA runtime and NVIDIA driver CUDA support in the error message, with a specific explanation when the PyTorch bundled runtime is newer than the installed driver supports.

## Why

Users installing Photon on L4s could get generic "CUDA unavailable" failures when pip selected a PyTorch CUDA runtime unsupported by the host driver. This makes that install/runtime mismatch actionable without adding any Photon-side CUDA-graph escape hatch.

## Validation

- `git diff --check`
- `uv run python -m py_compile kestrel/config.py tests/test_config.py`
- `uv run python -m pytest tests/test_config.py -q` was attempted, but local collection is blocked by the installed `kestrel-kernels` package lacking `attention.block_bidirectional_mask`, before these tests run.

### DIFF
--- a/kestrel/config.py
+++ b/kestrel/config.py
@@ -1,6 +1,7 @@
 """Runtime configuration objects for the Kestrel inference engine."""
 
 
+import ctypes
 from dataclasses import dataclass
 from pathlib import Path
 import re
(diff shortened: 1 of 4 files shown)


vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

vikhyat:
@codex review

PR #117, opened 2026-06-30
expose streaming for segment verb

ahepp:
The `segment` skill supports streaming, but this is not exposed by the engine. This commit wires it up similar to other skills.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ result = await stream.result()
 print(f"\n\nGenerated {result.metrics.output_tokens} tokens")
 ```
 
-Streaming is supported for `query` and `caption` methods.
+Streaming is supported for `query`, `caption`, and `segment` methods.
 
 ## Response Format
 
--- a/kestrel/engine/core.py
+++ b/kestrel/engine/core.py
@@ -752,19 +752,54 @@ class InferenceEngine:
             settings=settings,
         )
 
+    @overload
+    async def segment(
+        self,
+        image: Optional[np.ndarray | bytes],
+        object: str,
+        *,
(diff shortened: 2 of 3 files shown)


ahepp:
One note: I assumed that `pop_stream_delta()` was implemented as desired, which results in the output of the stream being the text `__BBOX__ <bbox> <path>`. We could also have the stream only return the path. Happy to add that onto this PR if desired.

```python
import asyncio

import numpy as np

from kestrel import InferenceEngine
from kestrel.config import RuntimeConfig

def make_image() -> np.ndarray:
    img = np.full((384, 384, 3), 255, dtype=np.uint8)
    yy, xx = np.ogrid[:384, :384]
    img[(xx - 192) ** 2 + (yy - 192) ** 2 <= 96 ** 2] = 0  # black circle on white
    return img


async def main() -> None:
    engine = await InferenceEngine.create(
        RuntimeConfig(model="moondream3-preview", enable_cuda_graphs=False)
    )
    stream = await engine.segment(make_image(), "circle", stream=True)

    async for update in stream:
        print(update.text, flush=True)

    await engine.shutdown()


if __name__ == "__main__":
    asyncio.run(main())
```
prints
```
__BBOX__{"x_min": 0.25, "x_max": 0.75, "y_min": 0.2509765625, "y_max": 0.7490234375}
M 0.357 0.993
 C -0.031 0.919 -0.031 0.253 0.318 0.04
 C 0.558 -0.031 0.9 0.084 0.974 0.317
 C 1.041 0.681 0.757 1.041 0.357 0.993
 z
```

ahepp:
@codex review

PR #120, opened 2026-07-06
moondream: wire whole-model decode megakernel into the engine

vikhyat:
## Wire the whole-model decode megakernel into the engine

Route eligible decode buckets through the whole-model megakernel (all layers in one persistent
CUDA launch) in place of the native `text_decoder`, for MD2 + MD3.

### The engine contract — a cheap gate, then a lazy run

The megakernel runs **eagerly** (a session build + one launch), which cannot happen inside a
captured CUDA graph. So the decision is a cheap **gate** the engine can ask *above* the graph
machinery, plus a **run** that builds lazily:

```python
tgt = kestrel_kernels.megakernel.deploy_target(model_name, device)   # (model_kind, arch, num_sms) | None
if tgt is not None and kestrel_kernels.megakernel.has_megakernel(*tgt, bucket):
    hidden = kestrel_kernels.megakernel.decode(          # eager, builds the session on first call
        model_name=..., text=..., text_config=..., device=..., page_table=...,
        max_seq_length=..., batch_size=bucket, embeds=..., batch_idx=..., input_pos=...)
else:
    hidden = text_decoder(...)                            # native (captured as before)
```

- `deploy_target` resolves the device to the `(model_kind, arch, num_sms)` of the variant that
  would serve it, or `None`. Cheap, builds nothing, never raises.
- `has_megakernel(model_kind, arch, num_sms, bucket)` is the cheap, no-build **SSOT policy
  predicate** — `True` when a shipped variant serves this bucket per the enable table. It builds
  nothing, launches nothing, and **never raises**: the never-raise-for-ineligibility property lives
  here now, on the gate.
- `decode(...)` returns the decoder's final hidden `[B, 1, H]`, building the per-bucket session
  lazily on the first call. Unlike the old `decode_or_none` it does **not** swallow failures into a
  `None` — it raises on a genuine internal error (build miss / kernel failure). The engine issues
  that first (building) call at warmup, and a raise there falls back to native (see below).

**All policy and the session lifecycle live in `kestrel_kernels.megakernel`** — the enable table,
deploy-SM selection, the lazy per-bucket build, the eager launch, the paged-KV args. The engine
carries **zero** megakernel policy.

### The capture fix — lift the decision above the graph machinery

The previous integration put the megakernel call *inside* `_run_decode_forward` — the exact
function the decode-graph machinery **captures**. Because the megakernel path is a session build +
eager launch (both illegal under CUDA-graph capture), during capture it produced *native*, so the
bucket **captured the native decoder and then replayed native forever**. With graphs enabled (the
production config) the megakernel never actually served.

This PR lifts the decision **above** the graph machinery, in `kestrel/runtime/decode_graph.py` and
the runtime's graph-build path:

- At graph-build time the runtime asks `has_megakernel` which buckets a shipped variant serves, and
  **warms** each one (the lazy session build, at the same warmup point capture used to touch).
- For every bucket that builds, it **skips native graph capture entirely** — no captured graph, no
  graph-pool memory — and routes that bucket's per-step decode to the eager `decode()` instead.
- Every ineligible bucket keeps the captured native path **byte-identically**.
- If a bucket's warmup build **fails**, it is excluded and captured native instead — the
  non-fatality is preserved as an engine-side warmup `try/except` around the gate, not a silent
  hot-path `None`.

Per-bucket decisions are independent, so a mixed fleet (e.g. B1 served by the megakernel, B8 native)
just skips capture for the served buckets and captures the rest.

### What this replaces

Deletes the old `kestrel/models/moondream/megakernel_decode.py` manager and the pre-capture
`prepare()` hook. The KV lifecycle lives with the kernels: MD3 reads/appends the engine's **paged KV
pool directly** (no seed, no private buffer), so a batch-composition change is just different
page-table rows next step. The engine diff is the gate + the capture-skip wiring.

### Validation

- **Capture-skip + routing** (`tests/test_decode_graph_manager.py`, CPU): an eager bucket is not
  captured and `run()` routes it to the eager forward; a native/excluded bucket still captures and
  replays; `clear()` drops the eager state. All green.
- **The gate on the real target** (H100 SXM, 132 SMs): `deploy_target("moondream3-…")` →
  `("moondream3", 90, 132)`; `has_megakernel(… , 1)` → `True`, `has_megakernel(… , 8)` → `False`;
  unknown model → `None`. The engine skips native capture for B1 and keeps native for B8, exactly as
  intended.

### Enabled today

MD3 and MD2 at batch 1 on H100 SXM (sm90, 132 SMs). Batches above 1 are deferred. Any other model /
GPU / bucket falls back to native (non-fatal).


vikhyat:
Revised per review, three axes (all re-verified on p2 / H100-SXM 132 SMs, real MD2+MD3 weights, production paged KV):

**1. Killed the JSON.** `megakernel_enable.json` → a typed `_ENABLE_TABLE` (dataclass `_EnableRow` rows) in `megakernel_decode.py`, reviewed like code. `num_ctas` selection folded into the manager (`_DEPLOY_SM_COUNTS` / private `_select_num_ctas`); the runtime no longer imports eligibility from the bridge.

**2. Slimmed the API.** Engine surface is now `prepare(bucket, batch_idx, input_pos)` in the decode-step prep + `decode(...)` in the captured forward (plus the existing `input_pos` write between replays) — nothing else. `seed_kv` is internal: `prepare` detects the prefill→decode transition itself (first decode / batch_idx change / input_pos discontinuity vs the manager's own session state) and seeds once; a steady contiguous step copies nothing (the cache self-extends across replays). Session build is lazy at the pre-capture warmup; LoRA routing is internal. The bench-parity `session.logits`/`forward_logits` moved off the shipped session to a bench shim (`mkl.benchmarks.decode_bridge_bench`).

**3. Shipped MD2.** Generalized to `WholeModelMd2BatchVM` (shared `_WholeModelDecode` base + `build_decode_session` factory). MD2 rebind inventory: cos8/sin8 + kv_lens8 (no TAU). `test_decode_bridge_md2_multistep` passes — byte-identity, capture==eager 8-step, native cross-check (step-0 cos 0.999895). MD2 B1 megakernel **wins 1.48x @seqlen16** (1200.85 vs 1777.25 µs native), **1.41x @seqlen512**, so MD2 B1 on sm90/132 is enabled.

**MD3 re-verified byte-identical through the refactor:** fingerprint cos 0.997372 / argmax 103; multistep (1) byte-identity (2) capture==eager (8- **and** 16-step, `[103]*N`) (3) native cross-check; rebind-gather wall +8.98 µs (< 15 µs budget); 128 captured replays clean. Fallback + kestrel import gates pass.

**KV seam, sized honestly** (ship doc §12): the VM read+append are already page-table-indirected — the seed copy is a KV **layout** mismatch, not an indirection gap. Cost = trivial wall (~97 MB one-time gather at history≈495, <1 µs/token amortized) but ~0.197 MB/token/row duplicated KV (~0.10 GB B1 / ~0.81 GB B8 at cap512, scales with context). Direct paged consumption is a specced deferred increment gated on HBM pressure.

The mkl-side changes (bridge factory + MD2 rebind/KV-seed + bench shim) are on the paired kestrel-proprietary branch.

vikhyat:
Memory update: the fp8-KV path is now wired end-to-end (dev + AOT bundle, 56-memref ABI, per-layer dequant scales; the bf16 pool gate is deleted). The KV pool halves exactly: ~21 GiB -> ~10.5 GiB on the ON leg. Combined with the engine-owned weight slab (#121/#384), the megakernel's peak-memory delta vs native shrinks from +21 GiB to ~+4-5 GiB. Real-checkpoint serve gates (multi-step + the 200-request accuracy parity) run next once the shared-host env restoration completes.

vikhyat:
## fp8-KV serve gate — real MD3 checkpoint, E2E through the engine (p2 SXM H100, 132 SMs)

With the bf16-KV gate retired (`ce3c495`) the whole-model decode megakernel now serves directly off the engine's production `torch.float8_e4m3fn` KV pool (fp8-aware read + native-byte-exact e4m3 append + per-layer `k_scale`/`v_scale` dequant). This is the real-serving-stack run the earlier random-weight benches could not gate. `moondream3-preview` HF checkpoint, `chartqa_eval.py` `--cot --max-batch-size 1`, temperature 0, prefix cache on. OFF = native `text_decoder` on the fp8 pool, CUDA graphs on. ON = megakernel bucket-1 on the **same fp8 pool**, eager. `verify_env.sh` KNOWN-GOOD on the base venv first.

### Token identity on the fp8 pool (4 sequential requests, native vs megakernel)

| req | native (ans / out_tok / reason) | megakernel (ans / out_tok / reason) | verdict |
|---|---|---|---|
| 0 | `15` / 12 / 34 | `15` / 12 / 34 | token-for-token identical |
| 1 | `3` / 10 / 23 | `3` / 9 / 27 | answer identical; CoT diverges (fp8) |
| 2 | `23` / 11 / 34 | `23` / 11 / 34 | token-for-token identical |
| 3 | `62` / 46 / 259 | `62` / 56 / 340 | answer identical; CoT diverges (fp8) |

All four final answers identical; 2/4 token-for-token; the two that diverge do so only mid-CoT and re-converge — the documented fp8 cos ~0.9968 class. **89 megakernel steps across 4 sequences, 0 native fallback, 0 fault** — the former cross-sequence step-3 crash (fp8/bf16 stride mismatch) is gone.

### N = 200 charts / 400 questions — ON vs OFF (warm, quiet node)

| Metric | OFF (native fp8-KV, graphed) | ON (megakernel fp8-KV, eager) |
|---|---|---|
| ChartQA accuracy | **76.75 % (307/400)** | **76.00 % (304/400)** — Δ 0.75 pp (within N=400 noise ~2.1 pp/σ) |
| Avg in / out tokens | 743.69 / 35.50 | 743.69 / 35.94 |
| Prefill throughput | 7097 tok/s | 7054 tok/s |
| Decode throughput | 338.81 tok/s | 340.85 tok/s (1.01×) |
| Requests / s | 9.54 | 9.48 |
| P50 / P90 / P99 (ms) | 166.93 / 379.58 / 588.83 | 160.41 / 371.00 / 605.35 |
| Decode steps mk / native / fallback | 0 / all(graphed) / 0 | **14001 / 0 / 0** |
| Peak GPU mem alloc / reserved | 17.530 / 17.627 GiB | 32.837 / 33.285 GiB |
| **KV pool (measured)** | **fp8 e4m3 6.000 GiB** | **fp8 e4m3 6.000 GiB** |

- **Accuracy parity** within noise; **throughput parity** (P50 slightly better on ON).
- **All-megakernel serve**: 14001 decode steps, zero native fallback, zero fault; one-time `cute.compile` 54.7 s.
- **KV memory**: fp8 pool is 6.000 GiB; the bf16-gate era forced a 2-byte pool = 12.000 GiB, so retiring the gate **saves 6.0 GiB of resident KV pool** at the default `kv_cache_pages` (scales 1:1 with pool size), read in place via direct paged-KV (no private KV copy). ON peak is higher than OFF only because of the resident whole-model weight stack (+12.1 GiB), orthogonal to KV dtype.

Verdict: the fp8-KV megakernel serves the real checkpoint at accuracy + throughput parity while restoring the fp8 KV memory saving the bf16 gate had cost, with no fallback and no fault. Ship blocker cleared. Full write-up in `mkl/docs/MEGAKERNEL_SHIP_INTEGRATION.md` §15.

vikhyat:
Superseded by #133. The replacement is rebased on current main, consumes canonical engine-owned decode metadata directly, consolidates the later integration fixes, and carries current H100/ChartQA evidence.

PR #124, opened 2026-07-11
moondream: route eligible decode buckets through the megakernel

vikhyat:
## Summary

- route eligible MD2/MD3 decode buckets through the eager whole-model megakernel above CUDA graph capture
- warm megakernel sessions before serving and keep native capture for ineligible or failed buckets
- use the production fp8 KV pool and session-owned page rebasing
- keep any runtime with an adapter workspace on native decode until the dynamic indexed-contraction ABI is wired

This supersedes the stale remote head of #120 and is rebased on current main after the engine-owned weight slab landed.

## Validation

- `python -m py_compile kestrel/models/moondream/runtime.py kestrel/runtime/decode_graph.py tests/test_decode_graph_manager.py`
- `git diff --check`
- Local pytest collection is blocked by the workstation CUDA torch install missing `libcudnn.so.9`; remote CPU and H100 validation are pending.

## Measured baseline

The pre-rebase #120 stack previously completed the 200-chart/400-question fp8-KV ChartQA gate on H100 SXM: 76.00% megakernel vs 76.75% native, 340.85 vs 338.81 decode tok/s, 14,001 megakernel steps, no fallback. This PR will be re-evaluated from the rebased head before review.


vikhyat:
Superseded by #133, which contains the current base megakernel engine integration on main with canonical engine-owned decode metadata.

PR #125, opened 2026-07-11
moondream: bind adapter workspace to indexed contractions

vikhyat:
## Summary

- stage logical adapter slots and physical routed-storage IDs as distinct per-row decode metadata
- allocate routed and dense weight families as fixed-address whole-model slabs while retaining native per-layer views
- maintain independent logical-slot rank tables for routed and dense contraction extents
- maintain the route-owned rank-major down-B layout at adapter load time
- expose the complete traced runtime tensor binding to `kestrel_kernels.megakernel.decode`
- serve adapter-enabled B1 through the megakernel while keeping larger adapter batches native until every row has emitted work
- add workspace storage-alias, dynamic-rank, and serving-gate tests

Depends on kestrel-proprietary #413 and #414 and is stacked on kestrel #124.

## Validation

- `python -m py_compile` on the modified engine/workspace/test modules
- `git diff --check`
- Remote CPU and H100 integration tests pending.

### DIFF
--- a/kestrel/models/moondream/decode_slot.py
+++ b/kestrel/models/moondream/decode_slot.py
@@ -49,7 +49,8 @@ class DecodeMetaBuffers:
             [
                 ("batch_idx", (max_batch_slots,), torch.int64),  # sequence batch idx
                 ("input_pos", (max_batch_slots,), torch.int32),  # token positions
-                ("lora_slot_ids", (max_batch_slots,), torch.int32),  # LoRA slots
+                ("lora_slot_ids", (max_batch_slots,), torch.int32),  # logical adapter slots
+                ("routed_storage_ids", (max_batch_slots,), torch.int32),
             ],
             device=device,
             pin_memory=True,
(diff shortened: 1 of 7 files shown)


vikhyat:
End-to-end integration update (p2 H100 SXM, 132 CTAs, FP8 KV, B1):

- 5fd24e9 removes the rank-zero session shortcut. An adapter-capable runtime now uses one dynamic-capacity megakernel session for both slot-zero and active adapter rows, matching per-sequence dynamic rank semantics and preventing a rank-zero session ABI from being reused with runtime tensors.
- Focused engine/runtime tests on p2: 62 passed.
- Full 400-question ChartQA legs completed with no megakernel warmup fallback.

| leg | accuracy | served mega launches | reported decode tok/s | aggregate decode ms/token |
|---|---:|---:|---:|---:|
| native base | 76.75% | 0 | 231.49 | 6.790 |
| mega base, dynamic capacity | 77.00% | 14,097 | 347.53 | 4.975 |
| native LoRA | 75.50% | 0 | 272.70 | 6.280 |
| mega LoRA | 76.00% | 13,368 | 295.42 | 5.748 |

Token-normalized speedups are 1.365x without an active adapter and 1.093x with the rank-16 adapter. The remaining 15.5% active-rank tax is a kernel/compiler performance issue, not engine fallback: launch counters prove both megakernel legs were served eagerly.

vikhyat:
Superseded by the later dynamic-adapter work and now parked while the generic dynamic-contraction runtime ABI is completed. The base integration is consolidated in #133; adapter wiring will return as one fresh PR against that shipped path.

PR #126, opened 2026-07-12
moondream: pin the MoE gate/up interleave in tests

vikhyat:
The 8-row gate/up interleave (#121) is a **shape-preserving row permutation**: gate row `n` lives at physical row `16*(n//8) + n%8`, up row `n` at that `+ 8`, and the per-channel scales and LoRA up-B carry the same permutation. Nothing can distinguish a correct interleave from an inverted, by-16, or dropped one except the index math — so an engine that stopped interleaving, or interleaved the weight but not the scale, would ship green and silently mispair every neuron.

**Nothing pinned it.** No test referenced `_interleave_gate_up_rows8` / `_interleave_up_b_rows8` at all, and `test_load_slot_copies_moe_weights` compared the workspace's `up_b` against the adapter's using `create_test_adapter`'s **constant** fill — which passes with the interleave, without it, or with an inverted one.

This is the same class of hole that let kestrel-kernels' warpgemv runtime ship a mispaired fused gate/up kernel (kestrel-kernels#409): the fixtures were in plain `[gate | up]` layout, so they could not see the layout at all.

## What this pins

- **The helpers**, with row-identifying values, asserted against the contract's index math directly — weights (with a hidden tail), scales (without one), and up-B (with its rank axis carried through).
- **The loader entry point** (`set_md3_moe_up_layer`), so a regression in the loader rather than the helper is caught end to end.
- **MoE LoRA up-B at workspace load**, with row-distinct values so the ordering cannot hide behind a constant fill, plus the standing `load_slot` assertion restated in terms of the interleave it actually expects.

Tests only — no production code changes.

## Verification

Mutation-tested, since a test that cannot fail is worthless here:
- stacking the interleave on the wrong axis (`dim=2` → `dim=1`) fails all four helper/loader tests;
- dropping the up-B interleave at workspace load fails the new LoRA test — while **every pre-existing test still passes**, which is precisely the gap.

`tests/moondream`: 64 passed, 1 skipped.

### DIFF
--- a/tests/moondream/test_lora_workspace.py
+++ b/tests/moondream/test_lora_workspace.py
@@ -3,6 +3,7 @@
 import pytest
 import torch
 
+from kestrel.models.moondream._moe_layout import _interleave_up_b_rows8
 from kestrel.models.moondream.config import TextConfig, TextMoeConfig
 from kestrel.models.moondream.lora import LoRA, TextLoRA, TextLoRAConfig
 from kestrel.models.moondream.lora_workspace import (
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

PR #127, opened 2026-07-13
moondream: route eligible decode buckets through the megakernel

vikhyat:
Rebased replacement for #124; the shared head is unchanged.

This preserves the original integration review unit: route eligible decode buckets through the eager whole-model megakernel, keep native graph capture for ineligible buckets, and cover the decode-graph handoff.

Verification:
- `PYTHONPATH=<kestrel-proprietary>/kestrel-kernels/python:$PWD .venv/bin/python -m pytest tests -q`
- 516 passed, 21 skipped

The dynamic adapter runtime is stacked separately.


vikhyat:

PR #128, opened 2026-07-13
moondream: bind adapter workspace to indexed contractions

vikhyat:
Rebased replacement for #125; the shared head is unchanged. Stacked on #127.

Adapter A/B weights remain ordinary indexed contraction inputs. The engine supplies fixed-address dense/routed slabs, logical and physical per-row selectors, and logical-slot rank tables. One dynamic-capacity session serves slot-zero and active-adapter replays without recompilation.

This refresh aligns the engine with the current `moe_lora_b_up_interleaved` ABI and deletes the unused streamed up-B companion. Rank-major down-B remains a measured physical layout required for coalesced route-owned down reads.

CPU gates:
- mixed rows use logical slots `[1, 2]`, routed storage `[0, 1]`, and different dense/routed ranks
- a later replay swaps rows and changes slot rank while every stable pointer/shape/stride/dtype signature remains unchanged
- `PYTHONPATH=<kestrel-proprietary>/kestrel-kernels/python:$PWD .venv/bin/python -m pytest tests/moondream/test_lora_workspace.py tests/moondream/test_moe.py -q`: 47 passed, 1 skipped
- full engine suite: 516 passed, 21 skipped

Existing p2 H100 ChartQA evidence from #125 remains applicable to the runtime design; GPU revalidation is still required for this rebased SHA.

### DIFF
--- a/kestrel/models/moondream/decode_slot.py
+++ b/kestrel/models/moondream/decode_slot.py
@@ -49,7 +49,8 @@ class DecodeMetaBuffers:
             [
                 ("batch_idx", (max_batch_slots,), torch.int64),  # sequence batch idx
                 ("input_pos", (max_batch_slots,), torch.int32),  # token positions
-                ("lora_slot_ids", (max_batch_slots,), torch.int32),  # LoRA slots
+                ("lora_slot_ids", (max_batch_slots,), torch.int32),  # logical adapter slots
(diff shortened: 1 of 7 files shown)


vikhyat:
Parking this stack while the generic dynamic-contraction runtime ABI is completed. The base integration is consolidated in #133; adapter wiring will return as one fresh PR against that shipped path rather than preserving this stale stack.

PR #129, opened 2026-07-13
build: gate megakernel runtime dependency

vikhyat:
Packaging prerequisite for #127.

`pyproject.toml` already requires `kestrel-kernels==0.4.7`, but the committed lock still installed 0.4.6. This refreshes `uv.lock` to the declared version and adds a fail-loud test for the runtime API #127 imports (`decode`, `deploy_target`, `has_megakernel`, `launch_count`).

**Blocked dependency:** the currently published 0.4.7 wheels were built before `kestrel_kernels.megakernel` landed. A clean `uv sync --no-editable` installs 0.4.7, then the new gate fails with `ModuleNotFoundError: kestrel_kernels.megakernel`. PyPI releases are immutable, so the paired wheel must be published under a new version; this PR must then update both `pyproject.toml` and `uv.lock` to that version before merge.

Current verification:
- `uv lock --check`: passes
- `.venv/bin/python -m pytest tests/test_kernels_megakernel_api.py -q`: intentionally fails against published 0.4.7, proving the deploy blocker

Do not merge #127 before this gate is green against the paired wheel.


vikhyat:

PR #130, opened 2026-07-13
moondream: keep decode padding mirrors in sync

vikhyat:
## Summary

- zero CPU host mirrors for padded decode rows alongside the existing GPU buffers
- pin the padding contract with a CPU regression test

## Root cause

The eager megakernel composition check consumes `batch_idx` and `input_pos` from CPU host mirrors to avoid a device synchronization. Power-of-two bucket padding zeroed only the GPU views, leaving stale live row/position values in those host mirrors. A long CoT run with a partially filled B8 bucket therefore interpreted padding as row 9 at position 749 while the GPU correctly used reserved row 0, and the reservation guard raised.

The paired kernels change in kp #435 treats reserved row 0 as padding for progression and extent checks. This engine change makes the host and GPU metadata agree at the source. Together they remove the false exception. Batch-composition changes still legitimately rebase the launch-local page map; that performance work is tracked in #435 and is not hidden by this correctness fix.

## Validation

- `tests/moondream/test_runtime_decode_padding.py` + `tests/test_decode_graph_manager.py`: 10 passed
- failure reproduced by MD2 ChartQA CoT with max batch 8 before the fix
- paired MD2 ChartQA CoT completed 1457 megakernel launches after the fix with exact native accuracy: 65.00% (260/400)
- final H100 bucket policy and full-path performance evidence are recorded in kp #435

### DIFF
--- a/kestrel/models/moondream/runtime.py
+++ b/kestrel/models/moondream/runtime.py
@@ -2386,7 +2386,9 @@ class MoondreamRuntime:
         slot.decode_coord_values[batch_size:graph_batch_size].zero_()
         slot.decode_size_values[batch_size:graph_batch_size].zero_()
         slot.meta.batch_idx.gpu[batch_size:graph_batch_size].zero_()
+        slot.meta.batch_idx.cpu[batch_size:graph_batch_size].zero_()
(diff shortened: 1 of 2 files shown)


vikhyat:

PR #131, opened 2026-07-13
Keep decode cohorts stable

vikhyat:
## What changed

- keep the first `max_batch_size` running requests as a stable decode cohort
- retain the one-request prefill headroom as a resident tail request
- promote that resident request only after a cohort member retires
- remove the per-token deferred-request rotation state
- add regression coverage for stable membership and temporary pipeline backpressure

## Why

The previous scheduler alternated which request was omitted whenever the running set exceeded the forward microbatch by one. That avoided starvation, but changed launch composition every token and gave affected requests irregular inter-token latency. It also forced launch-local decode state to track gratuitous batch membership changes.

The extra resident request remains useful for prefill headroom. Treating it as prefetched work waiting for a real cohort slot preserves that benefit while making decode cadence predictable for the active cohort.

The resident request can wait longer between its prefill-produced first token and its next token; this is the explicit latency tradeoff for preserving prefill-ahead without time-slicing the active cohort.

## Validation

- `ssh p1 'cd ~/code/kestrel && ~/.local/bin/uv run --no-sync pytest tests/scheduler -q'`
- 150 passed

### DIFF
--- a/kestrel/scheduler/scheduler.py
+++ b/kestrel/scheduler/scheduler.py
@@ -359,7 +359,6 @@ class GenerationScheduler:
         self._sampling_rng = torch.Generator(device=runtime.device)
         self._sampling_rng.manual_seed(torch.seed())
         self._pipeline = PipelineState()
-        self._last_deferred_request_id: int | None = None
         # Depth-1 spec overlap: the previously-launched macro-step's
         # (sequences, lazy SpecStepResult) awaiting commit while the GPU runs the
         # next step. None when no spec step is in flight.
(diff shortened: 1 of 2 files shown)


vikhyat:
@codex review

vikhyat:
Locked ChartQA A/B on `p1` (H100), holding `/tmp/kestrel_moe_bench.lock` for each full run.

Environment:

- full 2,500-question test split
- `moondream3-preview`
- `max_batch_size=4`
- prefix cache disabled
- exact latest nightly CPython 3.12 x86_64 wheel, SHA-256 `d02006281e1d4ce57e0d1dd68a28173eec76456f04b4d8860c67867c004c44cd`
- native `_pybridge.cpython-312-x86_64-linux-gnu.so` verified
- `uv run --no-sync` so the nightly wheel was not replaced by the lockfile pin

| Metric | Before (`ef4918b`) | After (`2a95e02`) | Change |
|---|---:|---:|---:|
| Requests/s | 38.86 | 39.30 | +1.13% |
| P50 request latency | 104.86 ms | 99.81 ms | -4.82% |
| P90 request latency | 161.32 ms | 160.97 ms | -0.22% |
| P99 request latency | 264.50 ms | 242.35 ms | -8.37% |
| Prefill throughput | 29,835.59 tok/s | 30,178.78 tok/s | +1.15% |
| Decode throughput | 128.82 tok/s | 130.17 tok/s | +1.05% |
| Accuracy | 84.08% | 83.80% | -0.28 pp |

The evaluator reports end-to-end request latency rather than per-token ITL. The stable-cohort policy nevertheless improves the request-latency median and tail while eliminating the known alternate-token cadence for active requests. The small accuracy difference is from changed batch composition/order under otherwise deterministic decoding.

vikhyat:
@codex review

PR #132, opened 2026-07-13
moondream: remove megakernel decode metadata mirrors

vikhyat:
## Summary
- stop maintaining CPU padding mirrors used only by the deleted megakernel composition detector
- keep the engine GPU metadata as the single decode input for native and megakernel paths

Stacked on #130. Pair with the kestrel-proprietary canonical decode ABI PR.

## Verification
- `compileall` and `git diff --check`
- local runtime test collection is blocked by the workstation missing `libcudnn.so.9`; GPU integration validation is tracked on the paired PR


vikhyat:

PR #134, opened 2026-07-14
moondream: simplify eager megakernel lifecycle

vikhyat:
## Summary

- resolve the immutable megakernel bucket set once when constructing the runtime
- lazily load the AOT megakernel session on the first real decode step
- keep megakernel buckets eager and avoid graph capture around a single AOT launch
- remove warmup launches, per-slot eager plans, failed-warmup state, and duplicate routing branches
- preserve per-step native fallback only for declaratively ineligible workloads; surface genuine runtime failures

## E2E evidence

H100 SXM, exact production `kestrel-kernels==0.4.8` AOT workflow wheel, ChartQA CoT=1, 200 charts / 400 requests per arm, prefix cache enabled, forward and reverse same-process A/B:

| Model / bucket | Native output tok/s | Megakernel output tok/s | Speedup | Accuracy |
| --- | ---: | ---: | ---: | --- |
| MD3 B1, FP8 KV | 335.02 | 414.51 | 1.237x | 76.50% vs 77.38%; repeat drift in both arms |
| MD2 B1, BF16 KV | 389.92 | 456.44 | 1.171x | 65.63% vs 65.50%; repeat drift in both arms |
| MD2 B2, BF16 KV | 626.56 | 714.42 | 1.140x | 65.00% vs 65.25%; native repeat drifted |

Every megakernel arm reported zero native fallbacks. MD3 routed every launch through B1/A1/U1; MD2 B2 routed every launch through B2/A2/U2.

## Verification

- `python -m pytest tests -q`: 539 passed, 4 skipped
- focused lifecycle tests: 16 passed
- private p2 environment gate: KNOWN-GOOD, including MD3 canonical fingerprint argmax 103 / cosine 0.999219

No package was published as part of this work.

### DIFF
--- a/kestrel/models/moondream/runtime.py
+++ b/kestrel/models/moondream/runtime.py
@@ -826,6 +826,18 @@ class MoondreamRuntime:
             )
             for slot_id in range(2)
         ]
+        # A shipped megakernel is already a single eager launch. Resolve its AOT
+        # buckets once and omit only their redundant native CUDA graphs.
+        megakernel_target = megakernel_decode.deploy_target(
+            self.model_name, self.device
+        )
+        self._megakernel_buckets = frozenset(
+            batch_size
+            for batch_size in make_decode_graph_batch_sizes(self.max_batch_size)
+            if megakernel_target is not None
+            and self._lora_workspace is None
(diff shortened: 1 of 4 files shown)


vikhyat:
@codex review

PR #136, opened 2026-07-22
Support query-specific stop tokens

ahepp:
Adds model-declared query stop tokens alongside the existing model EOS.

Query and flattened-chat states retain terminators in raw token results for metrics and logprob alignment, while excluding them from visible and streamed output.

Generated prefixes containing EOS or query stop tokens are rejected before replay.


ahepp:
@codex review

ahepp:
@codex review

ahepp:
@codex review

PR #137, opened 2026-07-23
moondream: pass host staging arrays to decode

ahepp:
## Summary

Update the Moondream runtime for the latest `kestrel-kernels` decode API:

- Use `batch_idx_host` and `input_pos_host`.
- Pass engine-owned NumPy views instead of CPU PyTorch tensors.

Tested end-to-end on H100 SXM with Moondream 3 ChartQA.


ahepp:
@codex review

PR #138, opened 2026-07-23
metrics: align prefill timing with first token readiness

vikhyat:
## Summary
- define prefill timing as scheduler admission through first model-token readiness
- define decode timing as first token readiness through request completion
- preserve TTFT as submission through first token, including queue time

This matches the phase boundary exported by vLLM (`first_token_ts - scheduled_ts`) and makes engine-side prefill throughput comparable without relying on client TTFT.

## Verification
- `153 passed` in `tests/scheduler` on p2
- `22 passed` across the focused metrics/executor/model-stream set
- `python -m py_compile` and `git diff --check`


vikhyat:
@codex review

vikhyat:
@codex review

PR #140, opened 2026-07-30
models: publish Qwen 3.5 inference support

vikhyat:
## Summary

- publishes production Qwen 3.5 image inference in `kestrel.models.qwen35`
- registers the validated under-40B dense and MoE checkpoint surface by Hugging Face repository ID
- preserves compiler-generated decode when the optional mkl runtime and matching AOT bundles are installed, with the public native path as fallback
- uses shared descriptor-driven generated decode, paged KV, preprocessing, tokenizer, rotary, and RMSNorm infrastructure
- keeps per-sequence M-RoPE deltas and indexed GDN state in runtime-owned state
- supports explicit local model and tokenizer paths without Hub access

Qwen support is intentionally image-only. Video, training, DFlash/speculative decode, compatibility facades, model-owned decode lifecycle, and unvalidated 122B/397B/Qwen 3.6 registrations are not included.

No manual tape or model-specific VM is added. The public Qwen package is 6,094 Python lines.

## Validation

- current kernels/native source/API stack with merged compiler: 119 passed, 7 skipped
- built public wheel has valid ZIP contents and explicitly declares the NumPy dependency used by Qwen image/runtime code
- prior p2 H100 same-process reasoning A/B: identical text, 16 sampled token IDs, and checkpoint fingerprint; performance remained within noise
- compiler tape shape, tape hash, argument-plan hash, binding keys, and carried-state contracts remained unchanged through the source cleanup
- git diff --check is clean

## Release gates

- finish kestrel-kernels 0.4.9 artifact build, then update the public pins to kestrel-kernels 0.4.9 and kestrel-native 0.1.6
- run the final real-checkpoint GPU test from that exact install

This PR remains draft and must not be merged before the final artifact test.


vikhyat:
Shared primitive adoption is now pushed in `f4b073e`:

- Qwen now imports generic `repeat_kv` and `default_inv_freq` from `kestrel.ops`; the duplicate Qwen implementations are deleted.
- The never-read, nonpersistent `original_inv_freq` runtime clone is deleted. The loader still intentionally ignores that possible upstream checkpoint key, so checkpoint compatibility and state ABI are unchanged.
- Qwen-owned source shrinks by 12 lines. The two new shared modules add 59 lines, for a temporary net +47 lines before Gemma rebases onto and extends the same primitives. Total Qwen package size is now 6,885 Python lines.

Validation on p2 H100:

- old/new text RoPE, vision RoPE, and grouped-KV tensors are bit-exact
- focused tests: 25 passed
- same-process alternating Qwen3.5-0.8B reasoning A/B: identical text, identical 16 token IDs, identical 384-key state hash (`c27c695e…`); 134.90 ms previous head vs 133.78 ms current (`0.9917x`, within noise)
- `git diff --check` clean

The PR remains draft and is not being merged.

vikhyat:
Shared RMSNorm ownership is pushed in `1b1c51a`:

- `kestrel.ops.norm.RMSNorm` now owns the scaled/unscaled module and delegates capability selection to `get_runtime().dense.rmsnorm`. No model or public width/AOT table is introduced.
- Qwen deletes its local RMSNorm function/class and duplicated kernel-width list; add-RMSNorm remains a distinct fused operation with its CUDA dtype/device/epsilon gate.
- Qwen loader offset folding now identifies the shared class. Parameter names, dtypes, state keys, and the gated RMSNorm are unchanged.
- The shared constructor preserves Gemma `with_scale=False` as an fp32, nonpersistent ones buffer. Generic tests cover Qwen/Gemma widths 128, 1152, 1536, and 5376.

LOC accounting (production source): Qwen `-26`, shared module `+35`, current PR net `+9`; Gemma can delete its 38-line duplicate on rebase, making the combined two-model result `-29` production LOC.

p2 H100 gates:

- 25 focused tests passed
- actual shared/runtime outputs are bit-identical at widths 128, 1152, 1536, and 5376
- same-process alternating Qwen3.5-0.8B reasoning A/B: identical text, 16 token IDs, and 384-key state hash; 134.41 ms previous head vs 133.96 ms shared (`0.9966x`, noise-level)
- `git diff --check` and compile checks clean

The PR remains draft and unmerged.

vikhyat:
Final artifact review now includes the exact source-free production stack on p2 H100:

- kestrel-kernels 0.4.9 + kestrel-native 0.1.6 import from the private wheel environment
- verify_env.sh reports KNOWN-GOOD; the packed sm90 generated-decode inventory resolves without kernel source
- Qwen3.5-2B generated decode is asserted present at B1/B2
- text answer 4, image answer red, concurrent answers 4 and 8

The generic production AOT resolver fix is merged as kestrel-proprietary #1262. This PR remains draft for owner review.

vikhyat:
Post-rebase source-free release gate: Qwen3.5-2B text, image, and concurrent C2 integration all pass on p2 H100 with `KESTREL_CUTE_JIT=0`. This run found and fixed the shared preparation-plan object/name bug in `9805679`; `tests/test_generated_decode.py` remains 3/3 green.

vikhyat:
Latest simplification is pushed in `70cfcfd`:

- deletes the model-level SDPA implementation, CUDA vision selector, and attention implementation environment/config axis
- renames the former `inference_ops.py` junk drawer to `gdn_state.py` and leaves only recurrent-state storage there
- moves vision metadata construction beside image preprocessing
- routes text and vision attention through the shared `get_runtime().attention.flash_attn_fwd` contract
- removes the newly added unit-test cases; the retained release gate is end-to-end text, image, and concurrent inference

The Qwen package is now 5,545 Python lines. Source-free p2 H100 validation with `KESTREL_CUTE_JIT=0`: text, image, and C2 integration all pass. Public Kestrel model code contains no SDPA call or attention-backend selector.

PR #141, opened 2026-07-31
Add generic media handoff to skills

ahepp:
Start down the path to supporting audio and video by allowing multiple and generic media to be passed to skills with the prompt.

Kestrel still only supports images at this time, and an adapter converts from the new media format to the legacy image format.

Tested with ChartQA on an L40, as well as local pytest.

The next follow up will be to have skills consume the new media format.


ahepp:
@codex review

PR #142, opened 2026-08-01
moondream vision: pad the patch embedding to the tensor-core alignment (588 -> 592) -- +86.6us on the served tower

vikhyat:
## What

`nn.Linear(588, 1152)` becomes `nn.Linear(592, 1152)` with four zero-filled input columns, and `create_patches` emits the matching 592-wide rows.

`588 = 14·14·3` is only **4-element aligned**. In bf16 Hopper `wgmma` operand descriptors need **8 elements (16 bytes)**, so at K=588 **no sm90 tile in cuBLAS's kernel set is eligible** and the heuristic falls back to an **Ampere `s16816`** kernel running on a Hopper part.

## Measured

Isolated GEMM, nc=13 (M=9477, N=1152), 1200 MHz, interleaved one-window A/B:

| | |
|---|---|
| Ampere `cutlass_80_tensorop_bf16_s16816gemm…align2` | **130.56 µs** |
| sm90 `nvjet_sm90_tst_192x144_64x5_1x2_h_bz_coopB_bias_TNT` | **36.96 µs** |
| | **+93.60 µs, 3.53×** |

**Served tower**, both arms running this same `vision_encoder`:

| | |
|---|---|
| K=588 tower (incumbent) | 21 476.6 µs |
| K=592 tower | 21 390.1 µs |
| **delta** | **+86.60 µs/forward** |
| disagreement / null bias | 0.001 pp / −0.097 pp |
| cos(K588, K592) | 0.999996 |
| clock | `locked@1200MHz`, 146 in-window samples, span 1200–1200 |

**93% of the kernel-level win survives into the tower.** The ~4–7 µs shortfall is the pad's own zero-fill and two extra launches (301→303).

## Exact, not approximate

The added columns are zero on **both** operands, so they contribute exactly zero to every dot product. A test asserts **equality — not a tolerance —** between the padded and unpadded linear **in fp64**, the dtype that can express the claim. (In bf16 the two bind different kernels with different accumulation order, so they agree to fp-noise; that is why the exactness argument is made in fp64.)

`592` rather than `608`: a measured sweep put their walls inside each other's spread, while 592 costs **+0.68%** extra FLOPs against 608's **+3.40%**.

## Real checkpoints load unchanged — no re-export

Checkpoints stay canonical at **588**. `_copy_patch_emb_weight` zero-extends on load for both the MD2 and MD3 formats.

It **raises** on any other mismatch rather than broadcasting, truncating, or padding on faith — this is the one place a real checkpoint could silently mis-load, and a wrong-shaped `patch_emb` yields a plausible-looking encoder that is quietly wrong. For MD3 the tensor is deliberately lifted **out** of the generic `weight_map` loop so that loop stays strict for everything else rather than growing a shape-tolerant path that would mask a genuine mismatch elsewhere.

## Design notes

**One owner for the width.** `aligned_patch_dim` is called by both `create_patches` and `build_vision_model`, so the activation and the weight cannot disagree about how wide the operand is.

**The pad is intrinsic to `create_patches`, not opt-in.** That function exists to produce the operand `patch_emb` consumes; an opt-in pad is a footgun that silently rebuilds the misaligned GEMM.

**The pad is zeroed on both operands.** `nn.Linear(592, …)` random-inits all 592 columns, so a freshly-built model carried garbage where the pad is — harmless *today* only because `create_patches` zeroes the matching activation columns, which is a **one-sided invariant**. `build_vision_model` now zeroes the weight pad too, so a built-only model matches a built-then-loaded one. Found by a bench guard refusing on live weights, not by reading the code.

**`create_patches` writes through a strided view** of the padded buffer rather than reshaping then slice-assigning: the permute makes the source non-contiguous so materializing costs one copy regardless, and this op moves ~11 MiB per forward at nc=13. Only the 4 pad columns are zeroed, not the whole buffer. The strided write is pinned by a test against a deliberately naive reference implementation.

## Tests

11 tests, all passing (plus the 66-test moondream suite, no regressions). They pin exactness (bit-equality by `torch.equal`, and fp64 equality end-to-end), the zeroed pad on both operands, the derived-not-hardcoded rule, and the checkpoint path including three refusal cases.

## Downstream

Full measurement write-up, including the served-tower A/B and the re-measured `native/TRT(fp16)` ratio (**1.0088 → 1.0054**, same pin, same engine artifact), is in kestrel-proprietary `mkl/docs/PATCH_EMBED_PAD_LANDED_2026_08_01.md`.

**Not touched:** the MPS/Metal vision path has its own implementation and was neither changed nor measured.


vikhyat:
Superseded by #143. This branch was cut from a local `main` that is 11 commits ahead of `origin/main`, so its PR diff carried eleven other commits' files (`runtime.py`, `skills/*`, `decode_graph.py`, two unrelated test files) that this change never touched. #143 is the same single commit cherry-picked onto `origin/main`, diff verified as exactly the three files this change owns. Closing rather than force-pushing.

PR #148, opened 2026-08-02
scheduler: revalidate prefill mode after binding

vikhyat:
## Summary

- derive the authoritative prefix-attention mode from the prepared sequence
- defer a row when reservation changes its mode after batch planning
- keep mixed attention modes out of a shared prefill launch

Prefix-cache reservations performed while binding earlier rows can evict a prefix that a later row expected to reuse. Revalidating after reservation closes that planning/binding race without adding a model-specific path.

## Verification

- `11 passed`: `tests/scheduler/test_launch_prefill_step.py`, `tests/scheduler/test_prefill_planner.py`
- tested against the `kestrel-kernels 0.4.9` nightly wheel

### DIFF
--- a/kestrel/runtime/state.py
+++ b/kestrel/runtime/state.py
@@ -103,3 +103,8 @@ class PreparedSequence:
     cache_result: _CacheLookupResult
     adapter_id: str | None
     image_hash: bytes | None
+
+    @property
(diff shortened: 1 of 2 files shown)


vikhyat:
Closing as redundant: the same scheduler fix merged in #146 while this branch was being validated.

PR #149, opened 2026-08-02
gemma4: fall back when generated decode is unavailable

vikhyat:
## Summary
- make bundled generated decode an optional Gemma acceleration
- select the smallest covering generated capacity for each active batch
- execute the same Gemma model through shared paged attention when no compatible generated program covers the batch

This keeps covered batches on their existing one-launch generated programs while allowing unsupported devices, missing bundles, and larger configured capacities to run correctly instead of failing engine construction. The shared paged correctness fallback fix is m87-labs/kestrel-proprietary#1326.

## Validation
- focused Gemma and registry tests
- generated and native decode paths completed the same end-to-end evaluation slice with matching aggregate accuracy
- syntax and whitespace checks

### DIFF
--- a/kestrel/models/gemma4/generated_decode.py
+++ b/kestrel/models/gemma4/generated_decode.py
@@ -27,7 +27,7 @@ def _rope_tables(runtime: Any) -> dict[str, tuple[torch.Tensor, torch.Tensor]]:
     }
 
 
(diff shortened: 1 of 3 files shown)


vikhyat:
@codex review

PR #153, opened 2026-08-03
engine: expose streaming segmentation

vikhyat:
## Summary

- propagate the public segment stream flag through the generic skill runner
- preserve the flag in SegmentRequest so the existing incremental SVG path stream activates
- add precise EngineResult/EngineStream overloads

This removes an API seam: the segment skill already implements incremental output, but InferenceEngine previously forced every segment request onto the non-streaming submit path.

## Verification

- uv run pytest tests/test_point_spatial_refs.py -q (8 passed)
- Python compile check
- git diff --check

This is intentionally a small Kestrel patch-release prerequisite for local segment streaming in moondream-python 2.0.

### DIFF
--- a/kestrel/engine/core.py
+++ b/kestrel/engine/core.py
@@ -726,19 +726,58 @@ class InferenceEngine:
             settings=settings,
         )
 
+    @overload
+    async def segment(
+        self,
+        image: Optional[np.ndarray | bytes],
+        object: str,
+        *,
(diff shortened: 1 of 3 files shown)


vikhyat:
Closing in favor of the existing generic ModelHandle capability path. ModelHandle already preserves the segment stream prompt and routes streaming through the shared skill runner, so no new engine-specific verb plumbing is needed.

PR #156, opened 2026-08-06
runtime: materialize finite batch domains

vikhyat:
## Summary

- exercise every admitted batch extent during dynamic compilation warmup
- preserve capacity as the first symbolic-positive compile/autotune input
- keep the helper target- and model-neutral
- prevent one-time per-shape setup from reaching serving requests

## B200 evidence

Controlled Gemma 4 vision-encoder runs used separate fresh Inductor caches on evee3.

- endpoint warm (`B1,B4`): 100.94 s initialization; first `B3` 51.09 ms; steady `B3` 11.94 ms
- full-domain warm (`B1,B4,B2,B3`): 93.00 s initialization; first served `B3` 12.58 ms; steady `B3` 12.58 ms
- the two intermediate warm calls add 103.02 ms inside the full-domain process
- both arms produced exactly two graphs and 206 cache files; unseen `B3` changed neither compiler counters nor cache size
- PyTorch 2.9 reproduced the same two-graph behavior; endpoint-warmed `B3` was 13.87 ms first-use versus 6.26 ms steady

The earlier multi-second observation is not attributed to intermediate-batch graph compilation here. This change only claims removal of the measured first-execution work, without graph or persistent cache growth.

## Verification

- `tests/test_runtime_compilation.py`: 1 passed on evee3
- `git diff --check`

### DIFF
--- a/kestrel/runtime/compilation.py
+++ b/kestrel/runtime/compilation.py
@@ -14,21 +14,27 @@ def materialize_dynamic_batch_domain(
     inputs_for_batch: Callable[[int], tuple[Any, ...]],
     synchronize: Callable[[], None] | None = None,
 ) -> None:
-    """Materialize the scalar-one and symbolic-positive batch regimes.
+    """Materialize every admitted batch before accepting serving work.
 
     ``torch.compile(dynamic=True)`` still specializes dimensions whose first
-    observed value is zero or one. A runtime that only initializes at batch one
-    can therefore compile the first grouped request in the serving path. Run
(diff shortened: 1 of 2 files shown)


vikhyat:
Closing after adding engine startup to the acceptance gate. Full-domain warming adds 103 ms to initialization to avoid roughly 38 ms of one-time B3 setup, with no graph, cache, or steady-state reduction. The earlier multi-second event did not reproduce as dynamic-batch compilation, so this is not a positive total-latency trade.

PR #185, opened 2026-08-27
Bind Qwen packed prefill topology and generated state transport

vikhyat:
## Summary
- bind packed Qwen GDN prefill to host-derived topology metadata
- enforce generated decode capacity and program contracts
- run fast sequence-major FP32 recurrence, then atomically commit validated rows into the BF16 value-major generated decoder pool
- keep native ReplaySSM seeding intact while generated prefill allocates no replay payload

## Validation
- focused Modal suite: 41 passed
- adversarial self-review: no remaining issues
- public main rebased at b943e051b6c253147c66cfcba68320a3fe1ff0b9

### DIFF
--- a/kestrel/models/qwen35/cache.py
+++ b/kestrel/models/qwen35/cache.py
@@ -43,6 +43,7 @@ class Qwen35InferenceCache:
         config: Any,
         paged_kv: Sequence[PagedKVCache | None],
         replay_capacity: int,
+        prepare_gdn_replay_state: bool = True,
     ) -> None:
         layer_types = tuple(config.layer_types)
         if len(paged_kv) != len(layer_types):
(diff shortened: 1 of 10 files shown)


vikhyat:
Closing as superseded after adversarial review. Current public main already writes generated packed-prefill recurrence directly into the compiler-selected BF16 value-major pool. This PR's remaining topology hooks target removed APIs, and its sequence-major FP32 temporary plus transpose/cast/index-copy would reintroduce the state conversion path we are removing. The one still-useful change, generated-artifact decode-slot capacity sizing, has been reimplemented as a focused current-main commit and integrated locally into PR #195 for validation. The native FP32 ReplaySSM fallback remains a separate explicit migration and is not being hidden by this closure.

PR #189, opened 2026-08-28
Pass Qwen decode row information to prefill

ahepp:
Qwen generated decode keeps recurrent state in rows allocated by the scheduler's page table.
Pass that information to prefill so the SM90 kernel can write each sequence's final GDN state directly into the correct decode row. Other prefill paths remain unchanged.


ahepp:
@codex review

PR #197, opened 2026-08-29
Add synthesis streaming interface

vikhyat:
## Summary

- add `ModelHandle.synthesize(...)` as a peer model capability
- let skills publish structured streaming output while preserving the existing text-delta contract
- support lossless compound streams for media such as incremental PCM
- let callers close generated streams or cancel single-pass runs without returning until admission preprocessing or in-flight GPU ownership is released
- add generic generated-runtime program selection, finalized weight-storage extension, and a fail-closed packed-sampling requirement

## Boundary

This PR contains the public synthesis/streaming interfaces and the generic engine, scheduler, and runtime support they require. It adds no model implementation and no TTS-specific admission, chunking, batching, codec, or media scheduling policy.

Autoregressive cancellation carries one generic event with the existing request, wakes the existing scheduler loop, and finishes through the existing admission, finalization, zombie-commit, release, and speculative-retire paths. Single-pass cancellation uses the same per-request signal and existing single-pass completion path: queued work is skipped, while launched work settles only after its device event completes. This adds no queue, registry, scheduler phase, model branch, or second lifecycle.

## Verification

- 191 focused engine, admission, scheduler, single-pass, capability-stream, and generated-runtime tests pass against the paired private runtime head
- the complete public suite passes on Modal: 1006 passed, 50 skipped; the four warnings are existing CPU-fallback warnings
- coverage includes closing a blocked iterator, cancelling and waiting for running admission preprocessing, prefill cancellation ordering, pause/shutdown settlement, and skipping queued or draining launched single-pass work
- the two legacy engine fixtures previously deselected from the H100 run now model the required runtime shape and pass without adding defensive production branches
- Ruff and `git diff --check` pass
- repeated adversarial API, stream-lifecycle, cancellation, speculative-decode, codec-cleanup, and scheduler-boundary review found no remaining user-facing issue or material simplification

This remains a draft and must not be merged before the paired private implementation receives final review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ extends precompiled execution across speech, Qwen, and Gemma workloads.
 - Added Gemma 4 26B-A4B base and instruction-tuned checkpoint support.
 - Added the model-bound `synthesize` capability and append-only streaming
   payloads for generated media such as PCM audio, including lossless compound
-  capability streams.
+  capability streams and explicit cancellation of generated streams.
 - Added Qwen3-ASR 0.6B and 1.7B and Parakeet TDT 0.6B v3 transcription for
   files and live PCM, with long-form chunking, progressive results, language
   and prompt controls, forced alignment, and word or character timestamps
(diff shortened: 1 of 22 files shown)


vikhyat:
Superseded by the current synthesis interface draft in #224. Closing the obsolete predecessor so the public API has one review target.

PR #207, opened 2026-09-02
Align MoE capacity with generated decode buckets

vikhyat:
## Summary
- map decode batches to the generated 1/2/4/8 MoE capacity buckets
- switch batches above 8 tokens to the existing prefill capacity policy
- cover Qwen3.5 and Gemma4 boundaries symmetrically

## Validation
- `tests/qwen35/test_moe.py`: 2 passed
- `tests/gemma4/test_moe.py`: 5 passed
- independent adversarial review: no findings

GPU resolution and fallback-clean behavior will be exercised by the H100 board retake.


vikhyat:
Superseded by the lower-level fix: Kestrel already passes the exact logical token count. The Kestrel kernels MoE dispatcher will map that request to the smallest calibrated physical capacity and reuse the corresponding handle. Keeping model wrappers free of backend bucket policy.

PR #238, opened 2026-09-10
fix qwen chat image accounting mismatch

ahepp:
In the past, image chat used an estimate of 4096 tokens for admission and kv cache allocation. PR #169 updated KV cache allocation to use the actual image size, but admission still uses the estimate.

As a result, if the actual size + gen limit is greater than the estimate, the scheduler stops generation after a single token.

To resolve, calculate the image token count from resize dimensions before preprocessing


ahepp:
@codex review

PR #247, opened 2026-09-18
Parakeet-TDT ternary (2-bit) student: moondream/parakeet-redux on CPU and MPS

vikhyat:
Draft for the owner's review; not to be merged without it. Depends on a kestrel-kernels release that ships `kestrel_kernels.ternary` and `kestrel_kernels._cpu` (kestrel-proprietary PR #3030); until then the ternary tests in CI skip.

## What
- `load_parakeet_tdt` detects a `ternary.json` manifest (local export dir or the `moondream/parakeet-redux` repo, pinned revision) and builds the ternary variant: `TernaryLinear` layers from `kestrel_kernels.ternary`, strict state-dict load, the export's separate q/k/v rows concatenated into the fused `qkv_proj`, the encoder's 1x1 convs routed through a helper that accepts runtime layers. Registered as `TERNARY_MODEL_ID` with a `device_types` restriction (CPU and MPS for now; CUDA requests are redirected with a warning).
- CPU defaults: the bf16 default resolves to bf16 only on CPUs with a native bf16 GEMM (AVX-512 BF16 / AMX) and to fp32 elsewhere; the depthwise conv runs as shifted multiply-adds in the [B, T, C] layout on MPS and on CPUs without oneDNN (torch's macOS path ran it one channel at a time, 53-67 % of the encoder's time on an M2).
- Engine: `RuntimeConfig.device` optional (`None` = the best device the model supports here), `resolve_model_device()`, `cpu_threads` / `ternary_mode` knobs, `needs_kv_pool = False` for ASR runtimes (no KV pool allocated), a CPU thread policy (physical cores capped at 8; 4 when a vnni layer is loaded), docs.
- Tests: depthwise path vs Conv1d+BatchNorm, dtype policy, device redirection, and an engine test over a tiny synthetic ternary export on CPU; a CPU-only CI job for the GPU-free subset.

## Verified (50 LibriSpeech dev-clean utterances, ternary rl6, transcripts identical on every device and mode, 0.76 % WER; teacher 0.51 %)
- EPYC 9575F: dense bf16 45.4x real time at 4 threads / 71.8x at 16; int8 GEMM mode 62-68x at 0.78 GB resident.
- M2 Air: MPS fp16 44-51x (0.7 GB process RSS), CPU fp32 16.0x, CPU int8 28.2x.
- All ASR surfaces through `InferenceEngine` on CPU and MPS: single and batched transcribe, word/char timestamps, long-form, progressive and live PCM streaming.

Follow-ups: fused Metal ops for the encoder (branch `ternary-mps`), the final weights and model card on the HF repo, a GPU regression run of the existing models before merge.


vikhyat:
Superseded by #253, which replays this branch onto main as one squashed commit (26 commits, including the two release-weight pins from this tip) followed by #248-#251 in order. Branch ternary-parakeet left intact.

PR #248, opened 2026-09-18
parakeet: relative-position attention through the conformer domain

vikhyat:
Stacked on #247 (`ternary-parakeet`). Pairs with kestrel-proprietary#3031.

Everything between the projections and `o_proj` — the relative shift, both score products, the mask and the softmax — is one runtime op now, the way the depthwise convolution already is:

```python
attended = get_runtime().conformer.rel_attention(
    self.qkv_proj(hidden), self.relative_k_proj(positions), self.bias_u, self.bias_v, mask, self.scale
)
```

It is the same op sequence, moved into `kestrel_kernels.conformer_ops`, so the CUDA and CPU backends keep their bits (a bit-identity test guards that in fp32 and bf16). The op takes the qkv projection fused rather than q, k and v separately, which is what lets a backend read the three in place instead of materializing the chunk/view/transpose chain the model used to build.

On an Apple GPU that block is 37 % of the encoder's torch dispatches, and the MPS runtime now has a Metal kernel for it. With the convolution kernel alongside it, on an M2 MacBook Air over the 50 dev-clean wavs:

| | dispatches / forward | encoder ms/utt | generate ms/utt | RTF on the 50 |
|---|---|---|---|---|
| torch reference | 2728 | 95.4 | 93.8 (51x) | 37.16 |
| both Metal kernels | 690 | 78.5 | 89.3 (54x) | 42.60 |

All 50 hypotheses are byte-identical to the stored reference, so the 0.76 % WER is unchanged.

`tests/asr/test_parakeet_mps_kernels.py` runs the encoder against whichever backend the host resolves to and against the torch reference and compares — on a Mac that exercises the Metal kernels, elsewhere it still guards the wiring — and asserts the module contains no `device.type` at all.


vikhyat:
Superseded by #253, which carries this commit unchanged on top of the squashed #247. Branch ternary-mps left intact.

PR #249, opened 2026-09-18
parakeet: no ternary weight knob, and a cheaper greedy decode step

vikhyat:
Stacked on #248 (`ternary-mps`). Pairs with kestrel-proprietary#3033, which carries the kernels.

## Measured on an M2 MacBook Air, 50 dev-clean wavs, best of 3

| | device memory | encoder ms/utt | generate ms/utt | greedy loop | RTF | hypotheses |
|---|---|---|---|---|---|---|
| before (`ternary-mps`) | 2495 MB alloc / 3524 driver | 80.4 | 101.5 | 35.6 ms | 41.0 | reference |
| this branch | **220 MB / 235 MB** | 87.4 | **90.6** | **27.8 ms** | 43.1 | **0 of 50 differ** |

Byte-identical, so the 0.76 % WER is unchanged.

## No weight knob

`RuntimeConfig.ternary_mode` is gone, `load_parakeet_tdt` loses the argument, and the bench script loses `--mode`. kestrel-kernels picks the one form worth shipping per device — packed on both shipped targets — and there is nothing left to select. The dequantized form stays as the test oracle and as what CUDA runs; it is not reachable from config.

## A cheaper greedy decode step

The greedy loop was a third of an utterance's wall clock and all of it dispatches and host waits. Two changes, both exact:

- the two argmaxes of a step go through the kernel runtime's `tdt` domain, which reads them once on MPS and twice on CUDA, where two measured faster end to end. The device-specific choice that used to be a comment in this file is now the backend's;
- the LSTM cell takes one `sigmoid` over the whole gate row rather than three over its quarters. Elementwise, so the bits are identical, and four fewer dispatches a layer.

A Metal kernel for the rest of the cell was tried in kestrel-kernels and rejected: it was faster but could not match MPSGraph's fp16 transcendentals, and in an autoregressive loop a one-ulp state difference derailed 2 of 50 utterances into repeated tokens.

No `device.type` anywhere in the module, as before.


vikhyat:
Superseded by #253, where its two commits are one (the first documented the weight modes the second removed). Branch ternary-metal-gemm left intact.

PR #250, opened 2026-09-18
parakeet: the encoder's glue through the conformer domain, and one relative-position projection

vikhyat:
Stacked on `ternary-metal-gemm` (#249). The model side of the fused CPU encoder in kestrel-proprietary#3034.

## What changed

- **The encoder block's glue goes through the `conformer` domain.** Every residual add feeds straight into the next LayerNorm, so the two are one op returning both tensors — the sum the next residual needs and the normalized input the next sublayer takes. The feed-forward activation and the convolution module's gate go through it too, as do the subsampling front end's two depthwise convolutions. The references are the ops they replace, so a backend without kernels for them is exactly where it was, and `alpha` is a power of two, so the bits are the ones the model had.
- **One relative-position projection for the whole encoder.** Every block projected the same table with its own weight, and the table does not depend on the activations, so all 24 are one `[2L-1, C] x [24C, C]` matrix multiply — 8.6 % of the encoder's projection work at a 24x wider output, and one activation quantization on the packed CPU path instead of 24. The blocks' weights concatenate by row into the encoder's single layer exactly as a block's q/k/v triple does, in the checkpoint and in the packed export alike.
- **The CPU picks its own cores.** How many was a number the caller had to guess, and guessing high was wrong: every kernel region here is short, so a hand-off across a last-level cache costs more than the extra cores return. The kernels size their pool to one cache domain and pin their workers there; `confine_to_cache_domain` puts this process on the same cores, because the thread that submits a region works in it.
- **The thread count comes from the config, not the environment.** `default_cpu_threads` no longer reads `OMP_NUM_THREADS`; `RuntimeConfig.cpu_threads` is how an operator names one, and naming one sizes the kernels' pool as well and suppresses the confinement — a caller who picks a count is managing placement themselves. The docstring records that a CPU deployment still wants `OMP_WAIT_POLICY=passive`, which is libgomp's variable rather than one of ours: with the encoder's work off torch, its idle OpenMP workers spin on cores the kernels' pool wants.
- The benchmark script applies the same policy when no thread count is given, so what it reports is what a deployment gets, and takes `--isa` for walking the compatibility instruction sets end to end.

## Numbers

50 LibriSpeech dev-clean utterances, batch 1, our scorer, bf16, `OMP_WAIT_POLICY=passive`, best of 3. Branch point and branch interleaved so both saw the same machine — an EPYC 9575F that is not idle (an 8-GPU distillation run with unpinned dataloaders started on it at 22:04).

| how the process is pinned | dense bf16 (removed) | int8 at the branch point | **this branch** |
| --- | --- | --- | --- |
| 4 cores | 52.4x | 77.9x | **83.9x** |
| 8 cores | 68.5x | 84.9x | **112.8x** |
| 16 cores (two CCDs) | 52.4x | 66.1x | **117.1x** |
| not pinned at all | 13.3x | 39.2x | **99.9x** |

The last row is the one a user hits by default, and the 16-core row is the policy working: handed sixteen cores across two CCDs it takes the eight that share a cache. Peak RSS 1.05 GB against 1.13 GB and 2.2–3.8 GB.

**M2 MacBook Air CPU, fp32**: 28.5x → **36.6x** on the default policy. Median latency 120 ms against 157 ms.

Transcripts unchanged: byte-identical to the recorded x86 reference, and on the M2 to the MPS reference (one utterance gains a trailing full stop the normalizer drops; WER 0.7585 % either way).

`tests/asr`: 44 passed / 8 skipped on the M2, 42 passed / 10 skipped on the x86 host.


vikhyat:
Superseded by #253: the encoder-glue commit is carried as is and the three thread-policy commits (one relative-position projection, the CPU picks its own cores, thread count from the config) are squashed into one. Branch ternary-cpu-fused left intact.

PR #251, opened 2026-09-19
Parakeet: pause-aligned long-audio segmentation (VAD head or energy pauses)

vikhyat:
## What

The Parakeet runtime cut long recordings into fixed 180 s chunks (`source.chunks(180)`). This replaces that with pause-aligned segmentation: cut the recording at pauses, never mid-word, every segment at most 30 s, full attention inside the segment, **no overlap**, text joined in order.

## Why — measured

`nvidia/parakeet-tdt-0.6b-v3`, six Earnings-22 calls of 15–22 minutes, word-level scoring through the TDT timestamps:

| protocol | WER |
|---|---|
| pause-aligned 30 s segments, no overlap (**VAD-head pauses**) | **6.37** (273 segments, median 26.5 s) |
| pause-aligned 30 s segments, no overlap (Silero pauses, reference) | 6.43 |
| pause-aligned 30 s segments, no overlap (**energy pauses**) | 6.81 (279 segments) |
| pause-aligned 30 s segments, ±4 s context margins | 6.60 |
| fixed 30 s grid, no overlap | 7.97 |
| whole recording, full attention | 10.42 |
| **fixed 180 s windows, no overlap — what this replaces** | **10.82** (959 deletions) |
| whole recording, local attention 512 / 256 / 128 / 64 | 10.39 / 11.35 / 11.59 / 12.70 |

Two readings behind the design:

- Fixed 180 s windows cost this model 4.7 points against pause alignment, mostly as deletions. Whole-recording inference (NeMo's default, local attention or not) is no better — it loses 4–6 points at every width.
- **Overlap is deliberately absent.** A transducer flushes a window's last words only when that window's audio ends, and deletes when a window opens mid-speech, so every context margin around a pause-aligned cut loses (±2 s / ±4 s / left-only 2, 4, 8 s were all worse; an emission-lag correction swept over 0.16–0.96 s had no effect). Margins recover about half the penalty of a *fixed* grid, but stay behind pause alignment.

## Cut policy

One walk from the start of the recording. The next cut is the midpoint of the **last** pause that ends before 30 s and begins after 1 s; with no such pause, the cut falls at the cap. A pause is a gap of at least 0.2 s between speech regions. Segments containing no speech at all are skipped rather than fed to the recogniser. Word timestamps are unchanged in kind — still derived from the TDT durations per segment and offset by the segment start.

## Pause source: a capability of the weights, not an option

`ParakeetTdtRuntime._speech_regions()` picks the pause source from the loaded tensors — one `if`. There is no environment variable, no request setting, and no bundled default head:

- **Weights carrying `vad_head.*`** use the checkpoint's own head: `Conv1d(1024→128, k=1) → SiLU → Conv1d(128→128, k=5) → SiLU → Conv1d(128→1, k=1) → sigmoid` on the encoder's convolutional subsampler output (213k parameters, fp16). Speech is runs of frames at p ≥ 0.5, gaps under 0.1 s bridged, runs under 0.1 s dropped. The subsampler is purely local, so the scan runs in 120 s blocks and never needs an encoder pass.
- **Everything else** — stock NVIDIA checkpoints, raw GPU Parakeet — uses frame energy: 20 ms frames in dB, noise floor at the 2nd percentile, speech at floor + 8 dB, gated on the 90th percentile clearing the floor by 8 dB. Pure numpy.

Both are just `Callable[[np.ndarray, int], list[tuple[float, float]]]`. The walk, the per-segment run, the text join and the timestamps are shared; only that callable differs. No new dependencies anywhere, and no device-specific code in the modeling path.

`weights.py` detects the capability before the strict load, so a checkpoint with a head loads it and one without is unchanged.

## Segmenting a forward-only source

`AudioChunks` decodes forwards only, so the walk runs against a carry buffer: 120 s blocks arrive, the pause source marks their speech, and a cut is emitted once more than 30 s are in hand. That cut is the one a walk over the finished file would make — every candidate pause lies inside the first 30 s of the carry, which is fully read by then. A pause still running at the end of the carry is usable too: whatever its true end, the midpoint measured so far is inside it, so the cut still lands in silence.

Live PCM windows are exempt: they arrive pre-cut, carrying decoder state and sample offsets into themselves, so they still pass through whole. The orchestrator's 180 s streaming block is now only the progress granularity and the live commit unit — the runtime cuts each of those at pauses before transcribing it.

## Simplifications (second commit)

The first pass had abstractions that only ever had one implementation. Removed:

- the `SpeechDetector` protocol, the `PauseSegmenter` class and the `VadHeadSpeech` wrapper → two module functions and one `if`;
- `head_speech`'s `device`/`dtype` constructor arguments → taken from the subsampler weight that consumes the features, so it follows the model wherever it was loaded;
- `ParakeetTdt.subsample`, a one-line wrapper over `encoder.subsampling`;
- `segment_edges` and `next_cut`'s `start` parameter, which production never used — the tests do that walk themselves.

## Adversarial review: what it found

Three real defects, fixed in the second commit:

1. **A cut could leave a remainder shorter than the 320 samples `parakeet_features` needs**, which would have raised "audio is too short to normalize" and failed the whole request. Such a remainder is now dropped. A recording that is *itself* that short still goes through, so it fails the way it always has rather than silently returning empty text.
2. **A yielded segment was a numpy view into the carry buffer** (`np.ascontiguousarray` does not copy a contiguous slice), pinning the whole ~10 MB buffer for as long as the runtime held the segment. Segments now own their memory.
3. **Segment starts accumulated float seconds per cut.** Now one division over a sample counter, so a long file cannot drift.

Checked and found correct, now pinned by tests:

- **Short, empty and silent audio.** A recording below one segment or below 1 s yields exactly one segment; empty audio still fails in `AudioChunks` as before; digital silence and a block with no dynamic range are *claimed* by the energy source rather than dropped on a guess.
- **Non-16 kHz input.** `AudioChunks` resamples to 16 kHz before any of this sees it; verified end to end from an 8 kHz source.
- **Long files.** A 6 h stream holds 52 MB regardless of length (carry ≤ 150 s plus one block); a 3 h file gives 400 segments covering it exactly, with no discontinuity and no segment over 30 s.
- **Block joins.** 20 synthetic recordings each crossing 2–3 joins produce cuts identical to the whole-file walk. The energy source reframes per block, so a pause edge can move by under one 20 ms frame — well under the 0.2 s minimum. The head's 5-frame conv sees zero padding at a join, so ±2 frames (160 ms) near each join have less context.
- **Device and dtype.** The head runs in the model's dtype — fp32, bf16 and fp16 all work on CPU, and under `torch.autocast` — with probabilities always returned as fp32. No `device.type` branch anywhere.
- **Malformed heads fail loudly.** A head the model has no room for, a partial head, and a mis-shaped head all raise through the strict load; none falls back to energy pauses. A checkpoint with no head loads strict and clean.
- **Timestamps across boundaries.** Skipping a speechless segment still advances the clock, so later segments keep their true place; starts are monotonic and the last segment ends exactly at the recording duration.
- **Batching.** Segments now vary in length *within* a batch, which fixed chunks never produced. Three concurrent requests share a batch and each keeps its own text, duration and clock.
- **Thread safety.** Nothing is cached on the runtime; the pause source is a module function or a fresh `partial`, and the segmenter is a generator with purely local state, so a shared runtime segments concurrent requests safely.

## Tests

`tests/asr/test_parakeet_segmentation.py` (12 tests) and `tests/asr/test_parakeet_longform.py` (3 added): the cut policy, the float tie where a frame-quantized 0.2 s gap evaluates to 0.19999999999998863, both pause sources, the head's post-processing, subsampler locality, capability detection and loud failure on malformed heads, the segmenter against the whole-file walk across block joins, silence skipping with the clock intact, short/offset/resampled recordings, the protocol end to end through the runtime with word-timestamp offsets, the live stream window passing through whole, and mixed-length batching.

`tests/asr` is green (42 passed, 8 skipped for CUDA). Full `tests/`: 1039 passed, 13 failed — all moondream / decode-path / generated-decode, and all 13 reproduce unchanged on `main` (`tests/qwen35` fails collection on `main` too).

**Cross-checked against the implementation the numbers above were measured with**, re-run after simplifying: on 20 synthetic recordings (280 segments, 260 cuts, each crossing 2–3 block joins) the pause lists are identical, the whole-file walk produces identical cuts (0 ns), and the streaming carry buffer matches the whole-file walk exactly (0 ns). The only difference is the final pause's *end* — whole frames there, true duration here, ≤ 19 ms, and never a cut.

## Caveats

1. **No real-weights run.** Validated by unit tests and by exact agreement with the reference implementation; not run against the real checkpoint (no GPU here and it is not in the local cache). Worth a Modal run before merge.
2. **120 s mel normalization.** `parakeet_features` normalizes mel statistics over whatever it is handed, so the head sees 120 s statistics. 120 s is the block size the 6.37 was measured with, but the head was *fitted* on single utterances; if this ever drifts, the block size is the knob.
3. **Energy pauses on mostly-silent blocks.** With speech in a small minority the 90th percentile sits in the noise, the dynamic-range gate fires, and the block is claimed as speech rather than skipped. That is the measured behaviour and the conservative direction, but only a checkpoint with a head skips long silences reliably.
4. **Encoder-graph buckets.** `_ENCODER_GRAPH_BUCKETS` tops out at 224 subsampled frames ≈ 17.9 s, so a 30 s segment (375 frames) falls back to eager — exactly as 180 s chunks did, so no regression, but adding a bucket that covers 30 s is now worth doing. Left out of this PR.

A companion change lets the ternary exporter ship a fitted head as `vad_head.*` (fp16), so a student checkpoint turns the head path on by itself. Round-tripped through this loader: capability detected, strict load clean, logits matching the fitted head to 7e-5 with identical speech decisions.

Draft — not for merge without review.


vikhyat:
Superseded by #253, where its two commits are one and the weights.py overlap with the ternary loader is resolved so a vad_head is attached on both loaders (a packed export shipping one would otherwise fail the strict load). Folded into the same PR for that reason rather than kept separate. Branch parakeet-vad-segmenter left intact.

PR #245, opened 2026-09-17
Add Qwen3.5 FP8 metadata and experimental DFlash serving

vikhyat:
## Summary

- Recognize the official Qwen3.5 block-FP8 checkpoint metadata and register `Qwen/Qwen3.5-27B-FP8`; reject explicitly unsupported FP8 formats.
- Add checkpoint-defined DFlash draft execution, verified-context caching, and an experimental single-sequence greedy serving adapter configured by `draft_model_path`.
- Support native recurrent continuation and accepted-prefix commit without repeating target dense projections. Independent captured recurrent layers share packed preparation and recurrence launches while retaining each layer's gate parameters and state ownership.
- Restrict the existing chat speculation cap to actual mask transitions rather than capping every ordinary chat step to one token.

Default autoregressive execution remains generated decode. The experimental speculative verifier uses native sequence-position kernels, not a generated T16 megakernel or batch-16 execution. Unsupported requests fail explicitly.

## Serving Measurements

Single B200, Qwen3.5-27B-FP8, the same DFlash checkpoint and three retained prompts, thinking enabled, greedy decoding, 15 drafts, 256 emitted tokens. Photon values are actual serving-engine decode measurements, not isolated kernels or a direct-loop harness.

| Prompt | Photon decode tok/s | Retained vLLM 0.29.0 DFlash tok/s |
| --- | ---: | ---: |
| Science | 253.34 | 248.15 |
| Code | 182.22 | 176.63 |
| Math | 319.55 | 325.40 |

Both decode rates exclude the first output token, which belongs to TTFT: 255 tokens divided by the decode span for these 256-token requests.

Before the fused-projection follow-up below, the private PR2992 thin-dispatch dependency removed duplicate validation and an unnecessary weight view while preserving launch validation. Its same-engine paired comparison improves decode by 1.030-1.034x over grouped preparation, preserving every output token and acceptance-round history across all 24 requests. All samples are retained, including a roughly 195 ms science-arm outlier; medians use three measured repetitions per arm after warmup.

Grouped preparation previously improved decode by approximately 1.039x over grouped recurrence alone, which itself improved serving decode by approximately 1.057-1.065x. Median science prefix-commit time falls from 4.60 ms to 2.98 ms to 1.76 ms across those two changes. Pre-fusion full-request wall-time medians were 1.084428 s, 1.487978 s, and 0.868615 s for science, code, and math respectively.

[...]

PR #252, opened 2026-09-20
Reuse speculative state banks and overlap finalization with drafting

vikhyat:
## Summary

- Allocate one spare recurrent-state cache at DFlash admission; alternate active
  and inactive banks on successful prefix commits. Ordinary cache callers keep
  the existing detached-snapshot behavior.
- Finalize accepted recurrent state on a dedicated stream. The next draft may
  proceed independently; target verification waits for the completion event.
- Order verification scratch reads, protect bank reuse, and wait before slot
  retirement/shutdown. Failed partial writes never publish the inactive bank.
- Reuse stream events instead of allocating an event per commit.

This PR is stacked on the existing Q27 FP8/DFlash branch (#245). Keep it open;
it is not authorization to merge that public feature branch or release.

## Validation

123 focused tests passed on B200, including real CUDA stream overlap and target
ordering, bounded repeated bank swaps, partial-copy failure, retirement ordering,
ordinary detached cache semantics, graph leases/recapture, and native recurrent
prefix checks. No local pytest.

Eight-B200 Q27 FP8 + DFlash diagnostic prompt, 256 output tokens, one cold and
two warm repeats. Same kernel archives between arms:

| Concurrency | Control aggregate request tok/s | Candidate |
|---|---:|---:|
| 1 | 350.29 / 351.91 | 362.08 / 363.97 |
| 8 | 1498.65 / 1503.65 | 1550.57 / 1556.04 |

Tokens and acceptance decisions match exactly for all three repeats in both
cells. Approximately 3.4-3.5% improvement on this prompt, not a full refreshed
board or a claim to beat vLLM. No arithmetic changes.

## Tradeoffs

The spare bank remains allocated for the admitted sequence lifetime. Q27 uses
about 75.75 MiB per extra bank, or 606 MiB at eight active requests. Verification
forks and other serving allocations remain; this removes destination tensor
allocation and cache/layer metadata copying from the packed commit path only.
Finalization still performs its actual state computation and copyback; overlap
depends on draft duration and device resource contention.


vikhyat:
Additional validation completed on B200: the full one-GPU C1/C2/C4/C8 run across three prompts (36 batches) matches the retained control token-for-token. Mixed prompt/length validation exercised every live cohort size from 1 through 8, repeated slot retirement/reuse, and also matched all control tokens. Allocated GPU memory after retirement was stable across all three candidate repeats (51,621,931,008 bytes); control was likewise stable. The steady-state host profile observed zero destination allocations and zero cache/layer metadata copies, with mean commit host time 0.391 ms versus 2.054 ms before this change. PR remains open and unmerged.

vikhyat:
Broader validation completed: full C1/C2/C4/C8 serving runs over three prompts and three repeats on both 1xB200 and the diagnostic 8xB200 path. The clean one-GPU rerun reproduces every prior output token. Each of the twelve eight-GPU cells has identical output tokens and accepted/committed counts across cold and warm repeats. Mixed-length retirement/re-admission coverage remains clean with no allocated-memory growth. The paired prompt-0 overlap gain remains approximately 3.4-3.5%; the full board includes other kernel changes and is not an isolated attribution to this PR. PR remains open and unmerged.

vikhyat:
Added graph-capture lifetime fix in 65b994ba; replay remains on the configured execution stream. The unchanged baseline reproducibly crashed after 240 requests because pooled capture-stream reuse invalidated a live cuBLAS workspace. A small reproducer confirms the aliasing failure.

B200 validation: 28 focused owned-stream, graph-cache, and draft-workspace tests passed. Eighty GEMM graph retirements returned allocations to baseline. A 576-request run completed with exact baseline tokens, followed by a 480-request randomized kernel A/B with exact tokens and acceptance vectors. Stream helper dependency is private PR m87-labs/kestrel-proprietary#3061; this public PR remains open and must not merge before the helper is available in the kernel dependency.

Staged-projection performance is separate and not included in this PR: randomized warm C8 request throughput 1629.83 -> 1674.13 tok/s (2.72%, including GC), prompt medians improve 1.5-1.7%. This is not a refreshed vLLM board or the campaign win target.

vikhyat:
Adversarial review found and fixed two issues in fddd66a1: (1) the capture helper was absent from the pinned released kernels package; stream ownership now lives in Kestrel runtime, with no new dependency, and (2) eager prefix tensors could be released while the finalization stream still read them; eager finalization now stays on the ambient compute stream while graph-owned buffers retain overlap. Re-review found no remaining issues. 118 focused tests passed on B200, covering graph/workspace lifetime, failure cleanup, stream uniqueness, repeated bank swaps, prefix finalization and decoder behavior. No local pytest. User explicitly authorized merging this PR after clean review; merge is into feature/q27-fp8-dflash only, not main or PR245, and is not a release.

PR #254, opened 2026-09-21
Batch speculative admission prefills with independent request state

vikhyat:
## Summary

Batch the target-prefill work inside speculative admission. Ready short prompts
share a forward pass; each request retains its own allocator slot, recurrent and
convolution state, DFlash context, first token, and failure handling. Verification
width is unchanged.

The scheduler collects only already-ready requests, bounded by free rows,
maximum concurrency, and a 512-token budget. A long prompt runs alone without
waiting. All backends use one admission-batch interface; the default preserves
independent admissions, while Qwen implements packed initial prefill. This is
not reuse of speculative verification caches, which require nonempty histories.
The budget is an internal scheduler limit, not a new configuration option.
Packed results populate the request's existing cache objects.

## Validation

- 277 tests passed remotely, including the entire scheduler directory, spec
  interface, and Qwen decoder suite. Covers existing
  scheduler semantics, mixed lengths, noncontiguous slots, invalid requests,
  duplicate owners, packed-forward failure, malformed callback results, and
  reserved-row cleanup. No local pytest.
- Fixed-prompt randomized admission A/B: eight serial requests ~199 ms, packed
  ~33 ms, including a mixed-length cohort. Both use the same new private source.
- Full-model isolation: changing only request 0, 3, or 7 in mixed-length C8
  batches leaves every other request's tapped features and all convolution and
  recurrent states bitwise unchanged. This checks ownership, not broad quality.
- 270 additional requests passed in a final strict-AOT serial/packed comparison
  across C1/C2/C4/C8. JIT disabled; compiler imports and fallback warnings fatal.
  All five generated projection plans cover rows 1--512 in eager and captured
  modes. Strict AOT reproduces the earlier JIT C8 outputs and round counts
  exactly. Earlier 180 mixed requests also exercise retirement and slot reuse.
- Greedy continuations are NOT bitwise identical: first divergence at token
  17--234 across the three homogeneous prompts. Final hidden/logit relative L2
  differences are ~11--12%, with first-token KL up to 0.0115. First tokens agree.
- Held-out paired likelihood: 56 cases, 16,238 reference tokens. Independent
  caches, identical continuation computation, only admission differs. A
  serial-versus-serial control is exact on 10 cases / 2,248 tokens.

| Domain | Cases | Tokens | Delta NLL | Paired bootstrap 95% interval |
| --- | ---: | ---: | ---: | --- |
| WikiText | 16 | 4096 | +0.002631 | [-0.000217, +0.005050] |
| WikiText 1024-token continuations | 8 | 8192 | +0.000530 | [-0.006445, +0.006790] |
| GSM8K reference answers | 16 | 2662 | +0.001382 | [-0.000880, +0.004252] |
| HumanEval reference solutions | 16 | 1288 | +0.000123 | [-0.001041, +0.001102] |

All upper bounds fall within the preselected +0.01 nats/token screening margin.
This is a likelihood screen, not task accuracy or maximum-context validation;
the long windows share one corpus. It does not claim bitwise numerical parity.

## Final B200 A/B

Same public source and scoped AOT bundle; control replaces only packed admission
with serial admission. Three prompts, 256 output tokens, DFlash15, one cold and
two warm cohorts per prompt. Throughput is total warm tokens / total wall time.

| Concurrency | Serial tok/s | Packed tok/s | Serial TTFT ms | Packed TTFT ms |
| --- | ---: | ---: | ---: | ---: |
| 1 | 361.55 | 361.34 | 22.05 | 21.95 |
| 2 | 665.84 | 670.14 | 44.85 | 24.82 |
| 4 | 1052.58 | 1056.62 | 88.35 | 27.03 |
| 8 | 1644.98 | 2046.78 | 177.88 | 33.82 |

C1 output is unchanged. C2--C8 continuations/verification rounds can differ;
GC remains included. The C8 1.244x is observed request throughput, not an
isolated kernel speedup. Against retained vLLM C8 1902.97 tok/s, the new result
is 1.076x; the campaign's 1.2x goal is not yet met. vLLM was not rerun.

## Release Integration

Merged into feature/q27-fp8-dflash after the final adversarial self-review and
another clean 277-test run, following explicit merge authorization. No release.
Release performance depends on the new kernel bundle containing the
merged private 512-row FP8 model/AOT extension; the currently published dependency
does not provide that larger-row fast path. Final validation used a newly built
workload-scoped packed AOT collection with current runtime source and the
installed native bridge, not a complete portable release wheel.

Implementation, adversarial self-review, held-out likelihood screening, and
strict-AOT serving validation are complete. Actual published-wheel validation
and the kernel dependency update remain release integration work. No package
release is part of this PR.


vikhyat:
Final adversarial self-review complete. Rechecked capacity accounting before publication, duplicate owners, partially failed allocation and packed forwards, session-construction failures, malformed batch results, first-token cleanup, independent KV/GDN indexing, recurrent-state ownership, and unchanged singleton semantics. No unresolved code findings. The simplified implementation passed the full 277-test scheduler/spec suite again on evee1. Existing held-out quality and strict-AOT serial/packed serving results remain applicable to this exact head. Merge is into feature/q27-fp8-dflash only, as now authorized; no release or version bump. The published kernel dependency still requires the newer 512-row bundle before release-performance validation.

PR #255, opened 2026-09-21
Defer verification state copies until commit

vikhyat:
## Summary

Keep full-block verification states in their graph-owned buffers until the
verification lease ends. Packed prefix finalization already reconstructs the
accepted state into owned destination banks, so exporting the rejected suffix
into temporary branch storage is redundant. No kernel or numerical change.

One cache ownership bit distinguishes borrowed graph outputs from ordinary
owned forks. Packed request rows retain disjoint views. The legacy full-block
commit detaches borrowed state before publishing it; eager owned full commits
remain unchanged. No runtime option, backend selection knob, or new dependency.

## Validation

- 314 focused scheduler/cache/graph tests passed remotely; four unrelated
  native continuation cases excluded from this focused run.
- Buffer-poisoning tests cover full and partial commits, original source
  isolation, graph-buffer reuse, and packed row ownership.
- 135 strict-AOT C1/C2/C4/C8 requests exactly match baseline output tokens and
  speculative acceptance/commit vectors. Compiler imports and fallbacks fatal.
- Another 48 strict-AOT mixed requests pass with 64/128/192/256-token limits,
  exercising early retirement, changing verification batch sizes, and slot reuse.
- Randomized same-process C8 A/B, 12 arms across three prompts: 288 requests,
  first two arms excluded from warm statistics. All outputs and acceptance
  vectors exact. Raw warm throughput 2018.66 -> 2055.88 tok/s (+1.84%), including
  GC. Baseline/candidate GC totals 411/289 ms confound that aggregate; low-GC
  prompt medians improve 1.1--1.2%. No time subtracted from throughput.
- Broader Qwen/scheduler suite initially gave 439 passes and 21 failures;
  unchanged baseline reproduces all 21 failures (438 passes). These include
  missing native fixture variants and unrelated integration tests; this is
  not a claim that the entire repository suite passes.

This supersedes the older deferred-state experiment on the current prefix
finalization path. It is an incremental improvement, not closure of the vLLM
campaign goal. No release/version bump.


vikhyat:
Adversarial self-review complete: checked graph-lease expiry, packed row isolation, unchanged eager ownership, full-block detachment before publication, partial-prefix reconstruction, inactive commit banks, and asynchronous finalizer lifetime. Buffer-poisoning coverage and 314 focused tests pass; 135 fixed-length requests reproduce exact outputs/acceptance and 48 mixed requests pass with strict AOT and fallback errors. Broader suite failures reproduce on the unchanged baseline. No unresolved change-specific findings. Merge remains confined to feature/q27-fp8-dflash; no package release.

